PR #3, opened 2016-04-17
Minor changes

rhart:
Some comments inline with the changes.


example/src/main/java/com/example/server/App.java
```diff
@@ -1,48 +1,39 @@
...
-                handler
-                    .get("hello", HelloWorldHandler.class)
-                    .all(ctx -> ctx.render("root")))
-    ).start();
+
+    RatpackServer.start(server -> server
+        .serverConfig(config -> config.port(serverPort))
+        .registry(Guice.registry(binding -> binding
+            .module(ServerTracingModule.class, config -> config
+                .serviceName("ratpack-demo")
```
rhart, line 23: I think it's better to encapsulate the `Brave` setup inside the module.

ratpack-zipkin/src/main/java/ratpack/zipkin/ServerTracingHandler.java
```diff
@@ -15,56 +15,8 @@
...
-    requestInterceptor.handle(requestAdapter);
-    ctx.getResponse()
-       .beforeSend(response -> responseInterceptor
-           .handle(responseAdapterFactory.createAdapter(response, responseAnnotationExtractor)));
-    ctx.next();
-  }
-}
\ No newline at end of file
+@FunctionalInterface
+public interface ServerTracingHandler extends Handler {
```
rhart, line 21: Breaking out an interface so users can override our default impl

example/src/main/java/com/example/server/App.java
```diff
@@ -1,48 +1,39 @@
...
-            .handlers(handler ->
-                handler
-                    .get("hello", HelloWorldHandler.class)
-                    .all(ctx -> ctx.render("root")))
-    ).start();
+
+    RatpackServer.start(server -> server
+        .serverConfig(config -> config.port(serverPort))
+        .registry(Guice.registry(binding -> binding
+            .module(ServerTracingModule.class, config -> config
```
rhart, line 22: Configuring the config like this is more ratpack-like.  Or using the config source support like yaml, env vars etc

ratpack-zipkin/src/main/java/ratpack/zipkin/ServerTracingModule.java
```diff
@@ -18,118 +18,105 @@
...
-    return config.spanNameProvider;
+  public ServerRequestInterceptor serverRequestInterceptor(final Brave brave) {
+    return new ServerRequestInterceptor(brave.serverTracer());
   }
 
   @Provides
-  public RequestAnnotationExtractor requestAnnotationExtractorFunc(final Config config) {
-    return config.requestAnnotationFunc;
+  public Brave getBrave(final Config config) {
+    Brave.Builder braveBuilder = new Brave.Builder(config.serviceName);
```
rhart, line 73: Really we need to be using `public Builder(ServerClientAndLocalSpanState state)` so we can override `ThreadLocalServerClientAndLocalSpanState`

example/src/main/java/com/example/server/App.java
hyleung, line 23: Yeah makes sense - guess we should also add something like `config.traceSampler(Sampler sampler)`, too. Think that would be all the things you can configure with `Brave`.

ratpack-zipkin/src/main/java/ratpack/zipkin/ServerTracingModule.java
hyleung, line 73: Sorry, I'm not super familiar with the Registry stuff - is there somewhere in Ratpack where I'd find an example of a similar component (i.e. something that stores/retrieves state from the registry)?

rhart, line 73: The Hystrix module is a good example.  See https://github.com/ratpack/ratpack/blob/master/ratpack-hystrix/src/main/java/ratpack/hystrix/internal/HystrixRegistryBackedRequestVariable.java#L36 and https://github.com/ratpack/ratpack/blob/master/ratpack-hystrix/src/main/java/ratpack/hystrix/internal/HystrixRegistryBackedRequestVariable.java#L39

See how much sense you can make of it, otherwise I'm happy to start something off or explain in more detail.

hyleung, line 73: Ok, I think I get the general gist of it - I didn't know you could get the current execution by doing `Execution.current()`. If I need to `add` multiple values of the same type (for example, there will be `Span` entries for both the current client and server spans) - how is that typically done? My initial thought was to wrap the value in some type but I was wondering if there was a better way.

rhart, line 73: Yep, wrapping in a new type is the best way.

hyleung, line 73: @rhart I've got an implementation in there now, as far as i can tell, it's working as expected. My initial attempt was actually sort of all messed up as I wasn't actually getting the _current_ execution every time - didn't see the bug until I threw a bunch of concurrent requests at it!

PR #6, opened 2016-04-23
Keep ClientLocalSpanState closer to the threadlocal version

rhart:
This is looking awesome!

I think though I would keep the ClientLocalSpanState impl closer to the threadlocal one for better traceability.  Unless there were reasons for the deviations I'm not groking.

I would also push the host and port into the module, we can get this from the ServerConfig.

### DIFF
--- a/ratpack-zipkin/src/main/java/ratpack/zipkin/internal/RatpackServerClientLocalSpanState.java
+++ b/ratpack-zipkin/src/main/java/ratpack/zipkin/internal/RatpackServerClientLocalSpanState.java
@@ -17,7 +17,6 @@ package ratpack.zipkin.internal;
 
 import com.github.kristofa.brave.ServerClientAndLocalSpanState;
 import com.github.kristofa.brave.ServerSpan;
-import com.github.kristofa.brave.internal.Nullable;
 import com.twitter.zipkin.gen.Endpoint;
 import com.twitter.zipkin.gen.Span;
 import ratpack.exec.Execution;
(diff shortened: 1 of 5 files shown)


ratpack-zipkin/src/test/groovy/ratpack/zipkin/internal/RatpackServiceClientLocalSpanStateSpec.groovy
```diff
@@ -115,6 +116,7 @@ class RatpackServiceClientLocalSpanStateSpec extends Specification {
             result.service_name == clientServiceName
     }
 
+    @Ignore
     def 'When no client service name set, should return null client endpoint'() {
```
rhart, line 120: I don't think this test is valid of the impl is like the threadlocal one.

ratpack-zipkin/src/main/java/ratpack/zipkin/internal/RatpackServerClientLocalSpanState.java
```diff
@@ -28,124 +27,113 @@
...
-    return registry.get()
-        .maybeGet(ServerEndpointValue.class)
-        .orElse(new ServerEndpointValue(Endpoint.create(serviceName, ip, port))).get();
+    return this.endpoint;
   }
 
   @Override
   public void setCurrentServerSpan(final ServerSpan span) {
-    registry.get().add(new ServerSpanValue(span));
+    registry.get().add(new CurrentServerSpanValue(span));
```
rhart, line 96: Possibly makes more sense for the types to be more specific

ratpack-zipkin/src/main/java/ratpack/zipkin/internal/RatpackServerClientLocalSpanState.java
```diff
@@ -28,124 +27,113 @@
...
     this.registry = registry;
-    registry.get().add(new ServerEndpointValue(Endpoint.create(serviceName, ip, port)));
+    this.endpoint = Endpoint.create(serviceName, ip, port);
   }
 
   @Override
   public Endpoint getClientEndpoint() {
     return registry.get()
-                   .maybeGet(ClientEndpointValue.class)
-                   .orElse(new ClientEndpointValue(null))
```
rhart, line 51: We can avoid these extra allocations

hyleung:
> I would also push the host and port into the module, we can get this from the ServerConfig.

Ah, I was going to ask if it was possible to pull the host and port in from somewhere else rather than having to set it in the config. ~~Can I get access to a `ServerConfig` instance in the Guice module?~~ (just had a closer look at the full PR)

PR #13, opened 2017-02-17
fix traced post method which was sending GET requests

edwardsoo-ss:


jeff-blaisdell:
👍

hyleung:
Oops, that was rather silly of me. Thanks for the fix @edwardsoo-ss!

PR #15, opened 2017-02-28
Made ZipkinHttpClientImpl extend ratpack's HttpClient interface, chan…

edwardsoo-ss:
…ged guice to use annotated binding


edwardsoo-ss:
hello @hyleung We thought it would be easier for users of this lib if the zipkin client just extends the ratpack http client interface. please take a look 
cc @jeff-blaisdell

jeff-blaisdell:
👍 I think this is a nice change @edwardsoo-ss.  Much easier to Zipkin enhance an existing project.

hyleung:
@edwardsoo-ss this looks great! I'll have a closer look at some point in the next few days and give it a spin. Also - still need to look into that PR build as well...

src/main/java/ratpack/zipkin/Zipkin.java
```diff
@@ -0,0 +1,11 @@
+package ratpack.zipkin;
+
+import com.google.inject.BindingAnnotation;
+
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+@BindingAnnotation @Target({ElementType.PARAMETER}) @Retention(RetentionPolicy.RUNTIME)
```
hyleung: Could we also allow field injection? Personally I'm not a fan of field injection but it might be a convenient thing for users.

hyleung:
@edwardsoo-ss are you seeing `CS`/`CR` annotations in Zipkin? Quite possible that I screwed something up but I'm not seeing them.

src/main/java/ratpack/zipkin/Zipkin.java
edwardsoo-ss: agreed

edwardsoo-ss:
@hyleung I forgot to update the get()/post() to use the traced request(), updated

jeff-blaisdell:
This isn't working for me either, even with the changes.

edwardsoo-ss:
updated to return stubbed/delegated body/headers/uri in the method capturing request spec to avoid some NPE in consumer code

jeff-blaisdell:
I've tested this and it works for me.  I really like how it will be pretty easy to move onto an interceptor based approach when that becomes available with this implementation.  Nice work @edwardsoo-ss

hyleung:
Sweet, works for me too :) Thanks so much, @edwardsoo-ss.

If I could ask a favour of you, could you rebase against master and push the branch again? I want to make sure I've fixed the PR build (I'm just doing `gradle clean check` for PRs and skipping the artifact signing).

edwardsoo-ss:
@hyleung of course, thanks for 👓 and testing

hyleung:
eh.... Coverage schmoverage...I'll update the README and my sample project when I get a few minutes and do a release. Thanks again, @edwardsoo-ss!

PR #18, opened 2017-03-05
Use standard annotation keys, add error status and handle NPE

llinder:
Zipkin defines several standard keys for paths and status codes.  Using these standard key names helps to keep things consistent with other tracers.

In addition Zipkin also defines an error annotation which colors spans in the UI when present.

Lastly there is an NPE when a client connection fails.  In this case the status object may be null and the tracer will throw an exception.

/cc @jeff-blaisdell @pfrybar @jterhune @edwardsoo-ss

### DIFF
--- a/src/main/java/ratpack/zipkin/RequestAnnotationExtractor.java
+++ b/src/main/java/ratpack/zipkin/RequestAnnotationExtractor.java
@@ -16,14 +16,20 @@
 package ratpack.zipkin;
 
 import com.github.kristofa.brave.KeyValueAnnotation;
+import java.util.Arrays;
 import ratpack.http.Request;
 
 import java.util.Collection;
 import java.util.Collections;
+import zipkin.TraceKeys;
 
 @FunctionalInterface
 public interface RequestAnnotationExtractor {
   Collection<KeyValueAnnotation> annotationsForRequest(Request request);
 
-  RequestAnnotationExtractor DEFAULT = (request -> Collections.emptyList());
+  RequestAnnotationExtractor DEFAULT = (request -> Arrays.asList(
+      KeyValueAnnotation.create(TraceKeys.HTTP_PATH, request.getPath()),
+      KeyValueAnnotation.create(TraceKeys.HTTP_METHOD, request.getMethod().toString())
+  ));
+
 }
--- a/src/main/java/ratpack/zipkin/internal/RatpackClientRequestAdapter.java
+++ b/src/main/java/ratpack/zipkin/internal/RatpackClientRequestAdapter.java
@@ -24,10 +24,12 @@ import com.github.kristofa.brave.http.HttpClientRequest;
 import com.github.kristofa.brave.http.SpanNameProvider;
 import com.github.kristofa.brave.internal.Nullable;
 import com.twitter.zipkin.gen.Endpoint;
+import java.util.Arrays;
 import ratpack.http.client.RequestSpec;
 
 import java.util.Collection;
 import java.util.Collections;
+import zipkin.TraceKeys;
 
 class RatpackClientRequestAdapter implements ClientRequestAdapter {
   private final RequestSpec requestSpec;
@@ -65,7 +67,10 @@ class RatpackClientRequestAdapter implements ClientRequestAdapter {
 
   @Override
   public Collection<KeyValueAnnotation> requestAnnotations() {
-    return Collections.singletonList(KeyValueAnnotation.create("http.uri", requestSpec.getUri().toString()));
+    return Arrays.asList(
+        KeyValueAnnotation.create(TraceKeys.HTTP_HOST, requestSpec.getUri().getHost()),
+        KeyValueAnnotation.create(TraceKeys.HTTP_PATH, requestSpec.getUri().getPath())
+    );
   }
 
   @Override
--- a/src/main/java/ratpack/zipkin/internal/RatpackHttpClientResponseAdapter.java
+++ b/src/main/java/ratpack/zipkin/internal/RatpackHttpClientResponseAdapter.java
@@ -17,10 +17,16 @@ package ratpack.zipkin.internal;
 
 import com.github.kristofa.brave.ClientResponseAdapter;
 import com.github.kristofa.brave.KeyValueAnnotation;
+import java.util.LinkedHashSet;
+import java.util.LinkedList;
+import java.util.List;
+import java.util.Set;
 import ratpack.http.client.ReceivedResponse;
 
 import java.util.Collection;
 import java.util.Collections;
+import zipkin.Constants;
+import zipkin.TraceKeys;
 
 class RatpackHttpClientResponseAdapter implements ClientResponseAdapter {
   private final ReceivedResponse response;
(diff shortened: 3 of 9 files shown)


hyleung:
@llinder looks good. Are those standard annotations part of a spec that I'm overlooking, btw? Not that I've necessarily been super diligent in that regard - for the most part I've been using Brave as a guide.

llinder:
@hyleung the "error" annotation is part of the Zipkin constants.  The others are just constants that are part of the Brave library.  AFAIK there isn't a written spec for the constants defined in the Brave library.  Besides being used in all of the Brave framework libraries they also appear to be adopted by other tracers like zipkin-go-opentracing.  https://github.com/openzipkin/zipkin-go-opentracing/blob/c2f58fe28dad57ea0daa99fe5089fdcf6ead60ad/_thrift/zipkinCore.thrift

src/main/java/ratpack/zipkin/internal/RatpackHttpClientResponseAdapter.java
```diff
@@ -31,13 +37,27 @@
...
+        annotations.add(KeyValueAnnotation
+            .create(Constants.ERROR, "error status " + httpStatus));
+      }
+
+      return annotations;
+    } else {
       KeyValueAnnotation statusAnnotation = KeyValueAnnotation
-          .create("http.responsecode", String.valueOf(httpStatus));
+          .create(Constants.ERROR, "missing or unknown status code");
+
```
hyleung, line 59: Perhaps a unit test for the `else` would be in order here as a way to indicate what is expected (i.e. an `ERROR` annotation, etc.) .

src/main/java/ratpack/zipkin/RequestAnnotationExtractor.java
```diff
@@ -16,14 +16,18 @@
...
 import java.util.Collections;
+import zipkin.TraceKeys;
 
 @FunctionalInterface
 public interface RequestAnnotationExtractor {
   Collection<KeyValueAnnotation> annotationsForRequest(Request request);
 
-  RequestAnnotationExtractor DEFAULT = (request -> Collections.emptyList());
+  RequestAnnotationExtractor DEFAULT = (request -> Collections.singletonList(
+      KeyValueAnnotation.create(TraceKeys.HTTP_URL, request.getPath())));
```
codefromthecrypt: probably more than path, right?

codefromthecrypt (commented):
very nice work! I appreciate the effort tweaking things to use standard keys, etc.

edwardsoo-ss:
looks good to me :+1:

jeff-blaisdell:
+1

src/main/java/ratpack/zipkin/RequestAnnotationExtractor.java
llinder: Thanks for flagging this.  I need to take a closer look at the result of `request.getPath()` from a running app to ensure this is the right key to use.  Something I have noticed from the Brave Servlet filter is that the `http.url` annotation is actual a path and not the entire URL. I'm not suggesting we should duplicate that here though :)

hyleung: @llinder did a little testing on my demo app - and also looked at the Javadoc for `ratpack.http.Request` - `getPath()` just returns the path portion of the URL, `getUri()` returns the path + query string.  That being the case, for `TraceKeys.HTTP_URL`, it would probably have to be more than just the path - as it's supposed to be _"The entire URL, including the scheme, host and query parameters if available"_

screen shot 2017-03-06 at 8 10 35 pm

~Oddly enough, that screen cap from above - I actually _removed_ the bit of code that added the `http.url` annotation. Not really sure where it's coming from!~ Found it: https://github.com/hyleung/ratpack-zipkin/pull/18/files#diff-d9a561d70ef7783a22f95f8b046fb2a0R69

codefromthecrypt:
heh yeah in brave slowly fixing this!
https://github.com/openzipkin/brave/blob/master/brave-http-tests/src/main/java/com/github/kristofa/brave/http/ITHttpServer.java#L235

src/main/java/ratpack/zipkin/RequestAnnotationExtractor.java
hyleung: Also noticed this in the Javadoc for `TraceKeys#HTTP_URL`
> This is optional as it may include private data or be of considerable length.

hyleung:
@llinder what do you think about separating out your NPE fix into a separate PR so we can get that fix in, and then continue massaging the annotations keys until we're happy?

src/main/java/ratpack/zipkin/RequestAnnotationExtractor.java
jeff-blaisdell: If you are looking for the full URL I think you want either `getUri` or `getRawUri`

https://ratpack.io/manual/current/api/ratpack/http/Request.html

jeff-blaisdell:
@hyleung Other than the value  for `TraceKeys.HTTP_URL` key was there some more you were looking for @llinder to change?

hyleung:
@jeff-blaisdell yeah, I think that was pretty much it (on tests/coverage etc. I don't mind backfilling those). Think @llinder wanted to take a closer look so I was just leaving time for that. What do you think about including the full URL *by default*? As I mentioned, in the Javadoc for TraceKeys, it does suggest that you may not want to include that information in some cases.

src/main/java/ratpack/zipkin/RequestAnnotationExtractor.java
llinder: I'm going to switch this to HTTP_PATH but feel free to request otherwise if you feel there is a good reason.

Based on experience I've found searching by path more meaningful and that including the host can make searching harder.  The reason for this is that most Zipkin storage components allow searching by value prefix.  If your host varies for any reason it makes searching by prefix a lot harder.

Having the query parameters in the key could be interesting though a bit scary from both a span storage perspective as well as a security perspective.  Hopefully most folks don't have giant query strings or sensitive data in their query strings but they do have a way of creeping in on occasion. 

Again just my 2 cents so please let me know if anyone feels otherwise.

llinder:
@hyleung If it helps I can split this up into 2 PRs.  I should have done that anyway so sorry for combining these things.  I've gone ahead and switched the annotation key to HTTP_PATH but if you think we need more discussion on this then I will definitely split out the NPE fix into a separate PR.

hyleung (approved):

hyleung:
@llinder going with path and method seems like a reasonable default (it's easy to override anyway). Thanks!

src/main/java/ratpack/zipkin/internal/RatpackClientRequestAdapter.java
```diff
@@ -65,7 +66,7 @@ public void addSpanIdToRequest(@Nullable final SpanId spanId) {
 
   @Override
   public Collection<KeyValueAnnotation> requestAnnotations() {
-    return Collections.singletonList(KeyValueAnnotation.create("http.uri", requestSpec.getUri().toString()));
+    return Collections.singletonList(KeyValueAnnotation.create(TraceKeys.HTTP_URL, requestSpec.getUri().toString()));
```
hyleung: @llinder given what we were discussing re. path vs. full URL - do you think we should go with path here as well?

src/main/java/ratpack/zipkin/internal/RatpackHttpClientResponseAdapter.java
llinder: Yes thanks!  Test added for these.

llinder:
I think the last update should cover the latest feedback.  I found that I missed updating the [RatpackHttpClientStreamedResponseAdapter](https://github.com/hyleung/ratpack-zipkin/blob/master/src/main/java/ratpack/zipkin/internal/RatpackHttpClientStreamedResponseAdapter.java) so that now has parity with the RatpackHttpClientResponseAdapter.

hyleung:
Sweet. Thanks, @llinder!

codefromthecrypt:
Applause

PR #21, opened 2017-03-28
If ratpack.server.ServerConfig#getAddress is null, set brave host/loc…

edwardsoo-ss:
…al endpoint to first site-local java.net.NetworkInterface#getNetworkInterfaces instead of loopback

@hyleung @llinder @jeff-blaisdell please take a look

### DIFF
--- a/src/main/java/ratpack/zipkin/ServerTracingModule.java
+++ b/src/main/java/ratpack/zipkin/ServerTracingModule.java
@@ -21,6 +21,8 @@ import com.github.kristofa.brave.http.SpanNameProvider;
 import com.google.inject.Provider;
 import com.google.inject.Provides;
 import com.google.inject.multibindings.Multibinder;
+import com.twitter.zipkin.gen.Endpoint;
+import ratpack.api.Nullable;
 import ratpack.guice.ConfigurableModule;
 import ratpack.handling.HandlerDecorator;
 import ratpack.http.client.HttpClient;
@@ -30,7 +32,9 @@ import zipkin.Span;
 import zipkin.reporter.Reporter;
 
 import java.net.InetAddress;
+import java.net.NetworkInterface;
 import java.nio.ByteBuffer;
+import java.util.Collections;
 
 import static com.google.inject.Scopes.SINGLETON;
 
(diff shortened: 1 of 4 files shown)


src/main/java/ratpack/zipkin/internal/RatpackServerClientLocalSpanState.java
```diff
@@ -37,24 +37,29 @@
...
-    this.endpoint = Endpoint.builder()
-                            .serviceName(serviceName)
-                            .ipv4(ip)
-                            .port(port)
-                            .build();
+    this.endpoint = endpoint;
     this.mdc = mdc;
   }
 
+  public RatpackServerClientLocalSpanState(final Endpoint endpoint) {
```
llinder, line 48: Are the other 2 public constructors defined here ever even used?  If not we should consider removing them or making them non public if possible.

hyleung: @llinder I think the `RatpackServerClientLocalSpanState(Endpoint, Supplier<MutableRegistry>, MDCProxy)` constructor is just used [here](https://github.com/edwardsoo-ss/ratpack-zipkin/blob/de80cbf2ad5551ac1b74f0bbb6fa2cac8ccbb901/src/test/groovy/ratpack/zipkin/internal/RatpackServiceClientLocalSpanStateSpec.groovy#L33) in a test setup. Could make it package-scoped.

[Update]
Actually, they're both used for tests. We could probably get rid of at least one of the extra constructors if we tweak the tests a little //cc @edwardsoo-ss

src/main/java/ratpack/zipkin/ServerTracingModule.java
```diff
@@ -115,6 +115,30 @@ else if (config.spanCollector != null) {
...
+    Endpoint.Builder builder = Endpoint.builder()
+            .serviceName(serviceName)
+            .port(port);
+    InetAddress address = configAddress != null ? configAddress : getSiteLocalAddress();
+    if (address.getAddress().length == 4) {
+      builder.ipv4(ByteBuffer.wrap(address.getAddress()).getInt());
+    } else if (address.getAddress().length == 16) {
+      builder.ipv6(address.getAddress());
+    }
+    return builder.build();
```
hyleung, line 128: Sort of a nitpick, i guess - but there's a couple of branches that aren't covered by tests. Coverage aside, it might be useful as a way of showing how the `ServerConfig` maps to the Brave stuff.

edwardsoo-ss:
removed a constructor, added some tests

llinder (approved):

hyleung:
Thanks, @edwardsoo-ss!

Issue #22, opened 2017-04-01
Upgrade to brave 4.1.0


drmaas:
At the minimum, this would involve using the new `brave.Tracer` instead of `LocalTracer`.

codefromthecrypt:
FYI Brave 4.1 will be out shortly and will include a replacement api for in-process propagation https://github.com/openzipkin/brave/pull/369

hyleung:
@adriancole yeah, I saw that this morning - might just skip straight to 4.1. Hoping to have some time to look into upgrading this repo soon (been busy with other stuff lately so haven't been doing the best job keeping up with changes).

hyleung:
Looks like @llinder is already on the upgrade to Brave 4.1 -  https://github.com/hyleung/ratpack-zipkin/pull/24

hyleung:
I've also started (in parallel) a spike exploring a more incremental approach - https://github.com/hyleung/ratpack-zipkin/pull/25

hyleung:
@llinder wanted to get your thoughts about how we should proceed (thought I'd do it here instead of in the PR).

At the moment, I'm thinking of maybe starting off with the more incremental approach - get this library updated to Brave 4.1.0 with as few major changes as necessary - that way we might be able to provide some features to address #10 because then we'll have `Tracer` at our disposal.  Then we can do that major version bump and start breaking shit 😛. 

If the timing works out, https://github.com/openzipkin/brave/pull/350 might be far along enough for us to incorporate. With that, plus your work on #24, we'll be able to drop the dependency on `brave-http` and maybe add support for some of the stuff @adriancole mentioned for 350.

llinder:
Sounds like a plan 👍

There are some nice things in #350 that is worth waiting for when its ready.  Until then it probably makes sense to go with #25 as you suggest.

hyleung:
@llinder looks like I jumped the gun a little - did more testing and realized that I was picking up an old snapshot artifact. So it wasn't actually working properly. It wasn't crashing, but it also wasn't recording any traces. I'd still like to be able to do something incremental but I'm a little less confident that we can pull it off now 😄

hyleung:
@llinder let's keep chipping away at #24 then, given that the incremental approach isn't really working out.

hyleung:
Updated to Brave 4.2 in #24

PR #23, opened 2017-04-03
Fallback to ratpack execution to get current spans

edwardsoo-ss:
this will allow propagating current spans to execution of ratpack's parallel batch

usage:

```java
List<Promise> promises = ...;
Span maybeLocalSpan = ...;
ServerSpan maybeServerSpan = ...;
return ParallelBatch.of(promises)
	.execInit(e -> {
		if (maybeLocalSpan != null) {
			e.add(new RatpackServerClientLocalSpanState.CurrentLocalSpanValue(maybeLocalSpan));
		}
		if (maybeServerSpan != null) {
			e.add(new RatpackServerClientLocalSpanState.CurrentServerSpanValue(maybeServerSpan));
		}
	})
	.yield();
```

@hyleung @llinder @Bijnagte @jeff-blaisdell take a look please

### DIFF
--- a/src/main/java/ratpack/zipkin/internal/RatpackServerClientLocalSpanState.java
+++ b/src/main/java/ratpack/zipkin/internal/RatpackServerClientLocalSpanState.java
@@ -56,10 +56,20 @@ public class RatpackServerClientLocalSpanState implements ServerClientAndLocalSp
 
   @Override
   public Span getCurrentClientSpan() {
+
     return registry.get()
         .maybeGet(CurrentClientSpanValue.class)
         .map(TypedValue::get)
-        .orElse(null);
+        .orElseGet(() -> {
+          if (Execution.isManagedThread()) {
+            return Execution.current()
+                    .maybeGet(CurrentClientSpanValue.class)
+                    .map(TypedValue::get)
+                    .orElse(null);
+          } else {
+            return null;
+          }
+        });
   }
 
   @Override
@@ -67,7 +77,16 @@ public class RatpackServerClientLocalSpanState implements ServerClientAndLocalSp
     return registry.get()
         .maybeGet(CurrentLocalSpanValue.class)
         .map(TypedValue::get)
-        .orElse(null);
+        .orElseGet(() -> {
+          if (Execution.isManagedThread()) {
+            return Execution.current()
+                    .maybeGet(CurrentLocalSpanValue.class)
+                    .map(TypedValue::get)
+                    .orElse(null);
+          } else {
+            return null;
+          }
+        });
   }
 
   @Override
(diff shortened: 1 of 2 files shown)


hyleung:
@edwardsoo-ss I'm not at all familiar with the `ParallelBatch` stuff, so could you elaborate on how this fallback logic works in that context? Does `Execution.current()` return something different when running in a `ParallelBatch`?

Also, I'm going to start looking at upgrading to Brave 4.x (see #22) - at the moment I don't know how that will look or whether it will change how this stuff will work.

edwardsoo-ss:
hello @hyleung, quoting https://ratpack.io/manual/current/api/ratpack/exec/util/ParallelBatch.html, a `ParallelBatch` is:
> A batch of promises to be processed, in parallel.
> 
> Parallel batches can be created via `of(Iterable)`.
> 
> Each promise will be executed in a forked execution. The `execInit(Action)` method allows each forked execution to be customised before executing the work.

The problem is that the forked executions do not get the same `ratpack.registry.Registry` as the execution context that started them, so we need another way to pass the trace context.

I have also tried invoking `ratpack.zipkin.internal.RatpackServerClientLocalSpanState#setCurrentServerSpan` and `ratpack.zipkin.internal.RatpackServerClientLocalSpanState#setCurrentLocalSpan` inside `ratpack.exec.util.ParallelBatch#execInit` after getting references to the server and local span at the point of the parallel batch creation, but the block supplied to `execInit` is not executed on a Ratpack-managed thread and the span setters will throw `ratpack.exec.UnmanagedThreadException` (at `ratpack.exec.internal.DefaultExecution#require`)

Another workaround is to get references to current spans and pass them to the promises to be executed in parallel using closure, but that requires the promises code to be aware of the tracing that is going on.

I don't think upgrading to brave 4.x will change this problem, since this is related to implementation of `ParallelBatch`. I can re-submit this suggestion/change until after the API upgrade if you prefer.

hyleung:
@edwardsoo-ss ah got it. So each `Promise` gets its own `Execution` via `fork`. And in the example you provided, you're passing the parent span into the closure?

edwardsoo-ss:
@hyleung yes, in the example I pass the spans to the `execInit` closure.
Currently in our projects I pass the spans to the promise using closures which is a bit different:
```java
Span maybeLocalSpan = ...;
ServerSpan maybeServerSpan = ...;
HttpClient httpClient = ...;

List<Promise> promises = Collections.singletonList(httpClient.post(uri, request -> {
    if (maybeLocalSpan != null) {
        localTracer.currentSpan().setCurrentSpan(maybeLocalSpan);
    }
    if (maybeServerSpan != null) {
        serverTracer.currentSpan().setCurrentSpan(maybeServerSpan);
    }
    ...
}));

return ParallelBatch.of(promises)
	.yield();
```

I also see that @llinder has done the same work for the 4.1 upgrade https://github.com/hyleung/ratpack-zipkin/pull/24 so I will just close this for now. thanks!

PR #24, opened 2017-04-04
WIP Brave 4.1 APIs

llinder:
Very very rough and untested attempt at using the Brave 4.1 APIs.

Next steps is to actually test this out and add abstractions where necessary.


src/main/java/ratpack/zipkin/internal/ZipkinHttpClientImpl.java
```diff
@@ -81,18 +75,16 @@ public void close() {
 
     @Override
     public Promise<ReceivedResponse> request(URI uri, Action<? super RequestSpec> action) {
-        return delegate.request(uri, tracedRequestAction(action))
-                .wiretap(receivedResponseResult -> responseInterceptor
-                        .handle(responseAdapterFactory.createAdapter(receivedResponseResult.getValue()))
-                );
+        final ExecuteWithSpan execution = new ExecuteWithSpan(tracer);
+        return delegate.request(uri, execution.execute(action))
```
codefromthecrypt: it might be useful to eagerly provision a span, then wrap delegate.request with `try(SpanInScope ws = tracer.withSpanInScope(span))` as that would allow anything invoking downstream see it (potentially helpful if SLF4J MDC is in use). You can lazy start that span in the execution thing..

codefromthecrypt: Key part is brave 4 spans are started independent of their creation exactly for this reason..

tracer.nextSpan(); for example, will create a span.. you can set that in scope independent of its timing, and setting it will allow others to read it (ex to add tags or read trace IDs).

src/main/java/ratpack/zipkin/internal/ZipkinHttpClientImpl.java
```diff
@@ -105,14 +97,64 @@ public void close() {
...
-            MethodCapturingRequestSpec captor = new MethodCapturingRequestSpec(requestSpec);
-            action.execute(captor);
-            HttpMethod capturedMethod = Optional.ofNullable(captor.getCapturedMethod()).orElse(HttpMethod.GET);
-            requestInterceptor
-                    .handle(requestAdapterFactory.createAdaptor(requestSpec.method(capturedMethod), capturedMethod.getName()));
-        });
+    private void tracedResponse(Result<?> result, StatusCode statusCode, @Nullable Span span) {
+        if (span != null) {
+            if (result.isError()) {
+                span.tag(Constants.ERROR, result.getThrowable().getMessage());
```
codefromthecrypt: potential NPE during the tagging.. 

I usually use something like..
```java
String message = throwable.getMessage();
if (message == null) message = throwable.getClass().getSimpleName();
```

src/main/java/ratpack/zipkin/internal/ZipkinHttpClientImpl.java
```diff
@@ -105,14 +97,64 @@ public void close() {
...
+        final Tracer tracer;
+        Span span;
+
+        ExecuteWithSpan(Tracer tracer) {
+            this.tracer = tracer;
+        }
+
+        Action<? super RequestSpec> execute(Action<? super RequestSpec> action) {
+            return (request -> {
+                Span parent = tracer.currentSpan();
```
codefromthecrypt: tracer.nextSpan() is a shortcut for this dance

codefromthecrypt (commented):
cool start! will be watching

src/main/java/ratpack/zipkin/internal/ZipkinHttpClientImpl.java
llinder: Thanks for the suggestion.  It wasn't clear to me if this would start a new trace if one wasn't already in progress.  I will switch to `tracer.nextSpan()` instead.

hyleung:
@llinder wow! I started working on the same thing last night - but clearly you're a lot further along than me. 😄

hyleung:
@llinder looks like there's a bunch of stuff that's going to be outright removed as part of this change. If that's the direction we're going in, then I suppose we might be looking at a major version bump (since there would be breaking changes)?

llinder:
@hyleung glad to hear you are looking into this as well!  Right now I'm just considering this a spike.  If the directions seems good then we can add parts back in that make sense for Brave 4.

My hope is that the changes will have minimal to no effect for most folks.  For anyone interacting with the Brave 3 APIs and wishing to upgrade there will definitely be some breaking changes.  We could always go a different direction and keep both Brave 3 and Brave 4 APIs in the project but that feels overly complicated.  My preference is to just use Brave 4 and do a version bump.  Then again I don't want to make decisions for other folks so hopefully we will get some feedback.

Also I don't want to feel like I'm taking over so feel free to steer this as you see fit and I will follow your lead :)

src/main/java/ratpack/zipkin/ServerTracingModule.java
```diff
@@ -15,107 +15,60 @@
...
-    );
-    if (config.spanReporter != null) {
-      braveBuilder.reporter(config.spanReporter);
-    }
-    else if (config.spanCollector != null) {
-      braveBuilder.spanCollector(config.spanCollector);
-    }
-    if (config.sampler != null) {
-      braveBuilder.traceSampler(config.sampler);
-    }
```
hyleung, line 114: I couldn't resist building your branch locally and giving it a spin - got an NPE due because I hadn't included a `Sampler` in my config. The way it currently is, it's optional - and I kinda like having a "batteries included" set of defaults so maybe that's something we can consider...

hyleung:
@llinder at the moment (we'll see how long it lasts) I'm trying to take an incremental approach but yeah, I have a hunch that over time a version bump with breaking changes might be the easiest. If it comes to that, them maybe we just get one more release out to get things synced up to the last 3.x Brave release (plus any PRs that might be relevant) leave it at that...

src/main/java/ratpack/zipkin/internal/RatpackCurrentTraceContext.java
```diff
@@ -0,0 +1,41 @@
...
+import brave.propagation.Propagation;
+import brave.propagation.TraceContext;
+import org.slf4j.MDC;
+
+public final class RatpackCurrentTraceContext extends CurrentTraceContext {
+
+  private final MDCProxy mdc;
+
+  private final TraceContext.Extractor<MDCProxy> extractor = Propagation.B3_STRING.extractor(MDCProxy::get);
+  private final TraceContext.Injector<MDCProxy> injector = Propagation.B3_STRING.injector(MDCProxy::put);
```
hyleung: What sorcery is this? 😁

codefromthecrypt: clever.. I never thought of doing it this way!

So, this is literally using the slf4j as the thread local storage for zipkin tracing. brilliant @llinder

codefromthecrypt: impact is that in your logging configuration, you would use the keys named same as headers (ex X-B3-TraceId). Some currently use lowercamel variants like traceId. changing the key format to something different than the header names might be requested in the future, but this is a reasonable default IMHO, and so little code compared to how I was thinking.

llinder: I think this can all be removed.  It was basically an attempt at using the MDC context for span context propagation.  It isn't really working as intended.  Will switch to the Ratpack MutableRegistry approach instead.

src/main/java/ratpack/zipkin/internal/RatpackCurrentTraceContext.java
```diff
@@ -0,0 +1,41 @@
...
+  }
+
+  @Override
+  public Scope newScope(TraceContext currentSpan) {
+    final TraceContext previous = get();
+    injector.inject(currentSpan, mdc);
+    return () -> injector.inject(previous, mdc);
+  }
+
+  private static final class MDCProxy {
```
codefromthecrypt: suppose proxy is needed as you can't get method reference to a static method?

src/main/java/ratpack/zipkin/internal/RatpackCurrentTraceContext.java
hyleung: ~I suppose you'd also be able to correlate your logs with Zipkin traces for free too!~ Doh! It could have been so beautiful!

src/main/java/ratpack/zipkin/ServerTracingModule.java
llinder: Thanks for flagging this.  Definitely should include defaults here!

src/main/java/ratpack/zipkin/internal/RatpackCurrentTraceContext.java
hyleung: This proxy thing is probably from my obsession about being able to test everything (iirc) - it's just some indirection to provide a seam for testing.

codefromthecrypt:
>
> I think this can all be removed. It was basically an attempt at using the
> MDC context for span context propagation. It isn't really working as
> intended. Will switch to the Ratpack MutableRegistry approach instead.
>
reminds me to make a base test class for CurrentTraceContext impl.  first
guess is that MDC would need some sort of dance to restore properly on
close.. anyway nice try very interesting!

https://github.com/openzipkin/brave/blob/master/brave/src/test/java/brave/features/log4j2_context/Log4JThreadContextTest.java

^^ has a dispatch approach which could still do log correlation, but uses a
more explicit means of attaching scope

codefromthecrypt:
>
> This proxy thing is probably from my obsession about being able to test
> everything (iirc) - it's just some indirection to provide a seam for
> testing.
>
heh powermockito is a sledge hammer I sometimes use (but it can become a
cure worse than the disease)

src/main/java/ratpack/zipkin/internal/RatpackCurrentTraceContext.java
llinder: Yeah thats exactly how I ended up with this :)  I don't think the MDC approach is the ideal approach with Ratpack.  It might be useful to put the trace context in MDC for other reasons but I think we can do without it for now.

src/main/java/ratpack/zipkin/internal/RatpackCurrentTraceContext.java
codefromthecrypt: :)

llinder: Log correlation would be a good reason to keep it :)  Easy to add back I suppose, still I seem to have a bug in here or Ratpack has a bug in their MDCInterceptor because it didn't seem to work.  I was probably just doing something stupid though ;)

llinder:
By the way, do people prefer I keep commit logs for all changes or is squashing ok?  I tend to do the later but having the history to look back on might help.

hyleung:
@llinder ¯\\_(ツ)_/¯  generally i'm pretty cool with whatever folks want to do on their own forks (tho' I do prefer a squash before the final merge). 

Having said that, if you're up for it, I'd love to be able to collaborate on this PR (even if it's just pitching in on tidying things up, writing tests, etc.) - in which case, obviously squashes on WIP is a a bit problematic.

llinder:
@hyleung cool I will avoid squashes to make collaboration easier.

src/main/java/ratpack/zipkin/TracedParallelBatch.java
```diff
@@ -0,0 +1,34 @@
...
+import ratpack.zipkin.internal.RatpackCurrentTraceContext.TraceContextTypeValue;
+
+public class TracedParallelBatch {
+
+  /**
+   * Add {@link TraceContext} propagation to {@link ParallelBatch#of(Promise[])}.
+   *
+   * Example usage:
+   *
+   * TracedParallelBatch
```
llinder: Pulled in the ParallelBatch work that @edwardsoo-ss was working on.  Essentially the same thing except I made some helper methods to abstract away the `.execInit` stuff.

src/main/java/ratpack/zipkin/TracedParallelBatch.java
```diff
@@ -0,0 +1,34 @@
+package ratpack.zipkin;
+
+import brave.propagation.TraceContext;
+import java.util.Arrays;
+import ratpack.exec.Promise;
+import ratpack.exec.util.ParallelBatch;
+import ratpack.zipkin.internal.RatpackCurrentTraceContext.TraceContextTypeValue;
+
+public class TracedParallelBatch {
```
llinder: Pulled in the ParallelBatch work that @edwardsoo-ss was working on. Essentially the same thing except I made some helper methods to abstract away the .execInit stuff.

edwardsoo-ss: :+1: thanks

src/main/java/ratpack/zipkin/internal/ZipkinHttpClientImpl.java
```diff
@@ -105,14 +109,63 @@ public void close() {
...
+    private Action<? super RequestSpec> actionWithSpan(final Action<? super RequestSpec> action, final Span span) {
+        return (request -> {
+            MethodCapturingRequestSpec captor = new MethodCapturingRequestSpec(request);
+            span.start();
             action.execute(captor);
-            HttpMethod capturedMethod = Optional.ofNullable(captor.getCapturedMethod()).orElse(HttpMethod.GET);
-            requestInterceptor
-                    .handle(requestAdapterFactory.createAdaptor(requestSpec.method(capturedMethod), capturedMethod.getName()));
+            span.name(spanNameProvider.getName(new DefaultRequestSpecNameAdapter(captor)));
+            span.tag(TraceKeys.HTTP_URL, captor.getUri().toString());
```
edwardsoo-ss: maybe use `ratpack.zipkin.internal.ZipkinHttpClientImpl.DefaultRequestSpecNameAdapter#getUri` just to be more consistent?

src/main/java/ratpack/zipkin/internal/ZipkinHttpClientImpl.java
```diff
@@ -105,14 +109,63 @@ public void close() {
         return request(uri, requestConfigurer.prepend(RequestSpec::post));
     }
 
-    private Action<? super RequestSpec> tracedRequestAction(final Action<? super RequestSpec> action) {
-        return action.append(requestSpec -> {
```
edwardsoo-ss, line 109: I didn't test this out, but I don't think the original action will be applied to the actual `RequestSpec` without this append/prepend

edwardsoo-ss: alternatively we can make `ratpack.zipkin.internal.MethodCapturingRequestSpec` modifies the actual `ratpack.http.client.RequestSpec` in its spec methods

src/main/java/ratpack/zipkin/internal/ZipkinHttpClientImpl.java
```diff
@@ -81,18 +79,24 @@ public void close() {
 
     @Override
     public Promise<ReceivedResponse> request(URI uri, Action<? super RequestSpec> action) {
-        return delegate.request(uri, tracedRequestAction(action))
-                .wiretap(receivedResponseResult -> responseInterceptor
-                        .handle(responseAdapterFactory.createAdapter(receivedResponseResult.getValue()))
-                );
+        final Span span = tracer.nextSpan();
+        try(Tracer.SpanInScope ws = tracer.withSpanInScope(span)) {
+            span.start();
```
edwardsoo-ss: this is called in `actionWithSpan`, probably not needed?

src/main/java/ratpack/zipkin/internal/ZipkinHttpClientImpl.java
```diff
@@ -81,18 +79,24 @@ public void close() {
...
 
     @Override
     public Promise<StreamedResponse> requestStream(URI uri, Action<? super RequestSpec> requestConfigurer) {
-        return delegate.requestStream(uri, tracedRequestAction(requestConfigurer))
-                .wiretap(streamedResponseResult -> responseInterceptor
-                        .handle(responseAdapterFactory.createAdapter(streamedResponseResult.getValue()))
-                );
+        final Span span = tracer.nextSpan();
+        try(Tracer.SpanInScope ws = tracer.withSpanInScope(span)) {
+            span.start();
```
edwardsoo-ss: ditto

src/main/java/ratpack/zipkin/internal/ZipkinHttpClientImpl.java
```diff
@@ -105,14 +109,63 @@ public void close() {
         return request(uri, requestConfigurer.prepend(RequestSpec::post));
     }
 
-    private Action<? super RequestSpec> tracedRequestAction(final Action<? super RequestSpec> action) {
-        return action.append(requestSpec -> {
-            MethodCapturingRequestSpec captor = new MethodCapturingRequestSpec(requestSpec);
+    private Action<? super RequestSpec> actionWithSpan(final Action<? super RequestSpec> action, final Span span) {
+        return (request -> {
+            MethodCapturingRequestSpec captor = new MethodCapturingRequestSpec(request);
+            span.start();
```
edwardsoo-ss: I think we should also annotate "cs" here

src/main/java/ratpack/zipkin/internal/ZipkinHttpClientImpl.java
```diff
@@ -105,14 +109,63 @@ public void close() {
...
+    }
+
+    private void resultWithSpan(Result<?> result, StatusCode statusCode, Span span) {
+        if (result.isError()) {
+            String message = result.getThrowable().getMessage();
+            if (message != null) {
+                span.tag(Constants.ERROR, result.getThrowable().getClass().getSimpleName());
+            }
+        } else {
+            int status = statusCode.getStatusCode();
```
edwardsoo-ss, line 141: annotate "cr" around here?

src/main/java/ratpack/zipkin/internal/DefaultServerTracingHandler.java
```diff
@@ -1,44 +1,90 @@
...
+
+    final Span span = (contextOrSamplingFlags.context() != null
+        ? tracer.joinSpan(contextOrSamplingFlags.context())
+        : tracer.newTrace(contextOrSamplingFlags.samplingFlags()))
+        .name(spanNameProvider.getName(new DefaultRequestSpanNameAdapter(ctx.getRequest())));
+
+    Request request = ctx.getRequest();
+    span.tag(TraceKeys.HTTP_URL, request.getUri());
+
+    ctx.getResponse().beforeSend(response -> {
```
edwardsoo-ss, line 48: missing "ss" here and "sr" below?

edwardsoo-ss:
I will try it out locally, looking good so far!

src/main/java/ratpack/zipkin/internal/ZipkinHttpClientImpl.java
llinder: cs/cr annotations are handled by the tracer.  Thanks for bringing it up though because I did forget to set the span kind to client here.

llinder:
Thanks for the comments and catching my mistakes @edwardsoo-ss!

hyleung:
@llinder Dude you're on 🔥! 

I was thinking about this a bit and it looks like you're on the path to pretty much on the path to doing away with `brave-http` altogether - i.e. getting down to the nitty gritty of annotating those spans yourself. What do you figure is the benefit of this approach over using `brave-http`'s APIs (all the adapters and whatnot)?

codefromthecrypt:
brave-http will mostly go away once more generic Client/ServerHandler
support is in
https://github.com/openzipkin/brave/pull/350/files#diff-e0f8554fb22e6ddd2bc65c6ea53965e2

If it remains, it would only include helpers related to http tags I think.

hyleung:
@adriancole ah, ic...so it would be a case of implementing (in this case) RatPack specific subclasses of these handlers? In any case, the logic around dealing with starting/finishing spans, making sure they're correctly annotated, etc. - that would reside in the base classes that are provided by `Brave` (e.g. `ServerHandler.handleRecieve`, `ServerHandler.handleSend`, right?

codefromthecrypt:
>
> @adriancole <https://github.com/adriancole> ah, ic...so it would be a
> case of implementing (in this case) RatPack specific subclasses of these
> handlers? In any case, the logic around dealing with starting/finishing
> spans, making sure they're correctly annotated, etc. - that would reside in
> the base classes that are provided by Brave (e.g.
> ServerHandler.handleRecieve, ServerHandler.handleSend, right?
>
The end goals of that PR are the following... not all would apply here

* be able to switch out B3 if needed (ex for internal things or another
format) that's the Propagation type
* be able to switch out the context implementation (even if ratpack
decorates, would be nice for folks to be able to weave in SLF4J etc)
* separate tagging implementation from tagging policy (ex be able to reduce
the tags emitted without knowing how they are created.. ex turn off
http.url or add a custom tag)
* have a single test framework that makes sure things mostly work
regardless of how they are implemented

So, once this is in, we'd have tests to say if you could add response.size
tag, for example, or to disable client address parsing regardless of
library.
I'm kindof meh about sharing base classes for this as some of the code
might be doable in another way.. sharing some config classes (especially
saying which propagation to use and what tags to record) might be helpful
regardless as that would allow someone a single switch.

what do you think?

src/main/java/ratpack/zipkin/internal/ZipkinHttpClientImpl.java
edwardsoo-ss: ah I see `brave.Span#kind` :+1:

llinder:
Thanks @hyleung but to his credit @adriancole is the one on 🔥 I'm just attempting to keep up with his breakneck pace :)   PR 350 looks like a great direction and I could see a lot of value using the pluggable span name, tags and address parsers here as well.

I mostly left `brave-http` out of the picture in this PR to get a better feeling for the core Brave 4 APIs without abstractions.  In general I found it way easier to use as apposed to Brave 3 and that the extra abstractions didn't provide as much value.  On the contrary, having a consistent way to handle span name, tag and address parsing seems very desirable, it just wasn't my initial focus.  Now that something has mostly been proved out with other frameworks I think it totally makes sense to apply them here as well.

hyleung:
@adriancole as always, thanks for the additional context :)

Re. `brave-http`, base classes, etc., for me the primary thing is being able to verify that a particular implementation has the correct behaviour. Especially at the beginning (hell, it's probably still the case now) when I wasn't super familiar with what needed to happen at a lower level, being able to implement a few adapters and interfaces and have it just plug in was handy. So having a way of verifying an implementation, whether it be that "single test framework" or otherwise would definitely be helpful.

src/main/java/ratpack/zipkin/internal/ZipkinHttpClientImpl.java
```diff
@@ -105,14 +106,64 @@ public void close() {
...
+    private void streamedResponseWithSpan(Result<StreamedResponse> response, Span span) {
+        resultWithSpan(response, () -> response.getValue().getStatusCode(), span);
+    }
+
+    private void responseWithSpan(Result<ReceivedResponse> response, Span span) {
+        resultWithSpan(response, () -> response.getValue().getStatusCode(), span);
+    }
+
+    private void resultWithSpan(Result<?> result, StatusCode statusCode, Span span) {
+        if (result.isError()) {
```
hyleung: @llinder I've been trying to write some tests for this class and can't for the life of me figure out how to get this branch to execute. Under what circumstances would we expect `result.isError()` to evaluate to true?

hyleung:
@llinder could I trouble you with one small thing on this PR before it's merged - can we split the `ParallelBatch` [stuff](https://github.com/hyleung/ratpack-zipkin/pull/24/commits/616eb34f80ed7d000ab75d5a7fb03613873fe8fc) into a separate PR? I had a question from https://github.com/hyleung/ratpack-zipkin/pull/23, plus I'd prefer not to have it mixed in with the Brave upgrade.

Other than that, let's get this merged in soon. What I'll do is probably one more 1.x release (for https://github.com/hyleung/ratpack-zipkin/pull/21), then bump it to `2.0.0-SNAPSHOT` and merge this in.

Does that sound reasonable?

[Update: I've gone ahead and done the bump to 2.0.0]

llinder:
Sounds good.  I will split the parallel batch stuff off.  I would like to get test coverage up a little more so I will try to tackle some of that this week as well.  Or we can address test coverage post merge, either way works for me.

src/main/java/ratpack/zipkin/internal/ZipkinHttpClientImpl.java
```diff
@@ -105,14 +106,64 @@ public void close() {
...
+    private void streamedResponseWithSpan(Result<StreamedResponse> response, Span span) {
+        resultWithSpan(response, () -> response.getValue().getStatusCode(), span);
+    }
+
+    private void responseWithSpan(Result<ReceivedResponse> response, Span span) {
+        resultWithSpan(response, () -> response.getValue().getStatusCode(), span);
+    }
+
+    void resultWithSpan(Result<?> result, StatusCode statusCode, Span span) {
+        if (result.isError()) {
```
hyleung, line 135: @llinder I wasn't able to write a test to exercise this code path - maybe I had the setup wrong?

hyleung:
@llinder I did have one [bit of the code](https://github.com/hyleung/ratpack-zipkin/pull/24/files#r112861075) that I had trouble writing a test for.

Other than that, I'm ok with addressing coverage after merge (I'm happy to help out with that).  I think the JaCoCo doesn't deal with lambdas very well so the coverage report is a bit misleading sometimes.

llinder:
@hyleung I removed the parallel batch stuff for now.  Also updated the dependency injection to use the Tracing component as suggested by @adriancole.  Lastly I updated the Ratpack request so that the tracing context headers are injected into the request.

Hopefully this is in a fairly good state to merge.  I will keep tracing Brave #350 work and incorporate new additions as they are released.

codefromthecrypt:
Hopefully this is in a good state to.. MERGED

Nice work and thanks for being the earliest into the new apis. I know it is
especially tricky with async.

hyleung:
@adriancole haha…to be clear, there are still a few bits an pieces (e.g. features that were removed) that I'd like to add back before we do a release - but we figured it would be better to make some forward progress.

@llinder I'm gonna add a couple of issues for the missing features (probably later today). If there's anything that you think needs to be included for an initial 2x. release feel free to add to the list (even if it's just a placeholder so we remember to discuss it) 😄

PR #25, opened 2017-04-05
Brave 4.x upgrade (spike to see if we can do this incrementally)

hyleung:
Still WIP but wanted to get some feedback on this early....


hyleung:
@llinder I did a quick test of this an as far as I can tell, I'm still getting traces like before. Basically my goal here was to change as little as possible. Because I'm lazy. 😛

src/main/java/ratpack/zipkin/internal/RatpackCurrentTraceContext.java
```diff
@@ -0,0 +1,52 @@
...
+  @Override
+  public Scope newScope(final TraceContext currentSpan) {
+    final TraceContext previous = get();
+    registrySupplier
+        .get()
+        .add(TraceContextHolder.class, new TraceContextHolder(currentSpan));
+    return () -> registrySupplier.get().add(TraceContextHolder.class, new TraceContextHolder(previous));
+  }
+
+  private static class TraceContextHolder {
```
hyleung, line 47: TIL that the registry's `get` and `maybeGet` don't actually allow null values (see. https://github.com/ratpack/ratpack/issues/1144) 😞

Hence this nonsense...

llinder:
Definitely a good iterative approach to maintaining existing functionality while introducing the Brave 4.x APIs.

src/main/java/ratpack/zipkin/internal/RatpackCurrentTraceContext.java
llinder: Thanks for the tip!

hyleung: @llinder crap, I don't think this actually works when you run actually run it in Ratpack because of where I initialize the `TraceContextHolder`. Because I did it in the constructor, I think it's being executed in a non-managed thread, which throws an exception.

src/main/java/ratpack/zipkin/ServerTracingModule.java
```diff
@@ -93,26 +101,29 @@ public ClientResponseInterceptor clientResponseInterceptor(final Brave brave) {
...
-        new RatpackServerClientLocalSpanState(
-                buildLocalEndpoint(
-                        config.serviceName,
-                        serverConfig.getPort(),
-                        serverConfig.getAddress()
-                )
-        )
-    );
+  public Brave getBrave(final Tracer tracer) {
+    return TracerAdapter.newBrave(tracer);
```
hyleung, line 105: @llinder I don't think this works the way I thought it would - looking at the implementation, it appears to construct a new `Brave` using an instance of `ThreadLocalServerClientAndLocalSpanState`, which obviously isn't much help to us. I had hoped that it would use the `CurrentTraceContext` instance that's already present in the `Tracer`,  but no 🎲

codefromthecrypt:
It might be a lot of work and probably special hooks to attempt to emulate
the old thread state model (server-client-local) with any alternative due
to coordination needed to tell if whatever current is a server client etc
or not. It is further difficult because the threadbinder code needs to work
on a concrete span type.

Personal plan is to defer that sort of effort in favor of completing brave
4 api instrumentation. That way new installs can be simple.

If after that I can make time without killing something else, I could help
try to make some sort of syncrhonizing machinery to adapt the threadbinders
to current span (it likely needs to also read braves internal state or keep
its own to know which is a client server or not). If anyone else can help
with that, I'd be happy to have it!

llinder:
@adriancole I agree and would rather see the brave 4 APIs progress as apposed to figuring out the mapping between `CurrentTraceContext` and `ServerClientAndLocalSpanState`.   Personally I don't think the company I work for is in a situation where we _need_ to mix brave 3 and 4 APIs in a single project.  My plan is to instead move from Brave 3 to 4 on an app by app basis and avoid the TraceAdapter entirely.

src/main/java/ratpack/zipkin/ServerTracingModule.java
llinder: Can you use `public static Brave newBrave(Tracer tracer, ServerClientAndLocalSpanState state)` instead?  That would allow you to use `RatpackServerClientLocalSpanState` instead of the thread local one.

src/main/java/ratpack/zipkin/internal/RatpackCurrentTraceContext.java
llinder: Bummer :(  Looking at things there doesn't seem to be a good way to use the `TraceContextHolder` with Brave 3 without some extra mapping logic as @adriancole suggested.  Not sure how much work would be involved with building a bridge between `TraceContextHolder` and `ServerClientLocalSpanState` but I suspect it wouldn't be trivial.

hyleung:
@adriancole @llinder yeah, I think I'm gonna close this PR/spike. If it was a quick win and we could keep compatibility then that would be been nice, but I agree we should focus on the Brave 4 upgrade.

PR #30, opened 2017-05-09
Spikes towards using http adapters

codefromthecrypt:


codefromthecrypt:
fyi @llinder I tried in the second commit to pass an atomic ref around instead of a span. This seemed to untangle some things about lazy init, although not everything.

src/main/java/ratpack/zipkin/internal/MethodCapturingRequestSpec.java
```diff
@@ -60,7 +67,8 @@ public RequestSpec headers(Action<? super MutableHeaders> action) throws Excepti
 
   @Override
   public RequestSpec method(HttpMethod method) {
-    capturedMethod = method;
+    this.capturedMethod = method;
+    span.set(impl.handler.handleSend(impl.injector, getHeaders(), this));
```
codefromthecrypt, line 71: basically I'm looking for an "onRequestSent" callback, and maybe.. this is it?

src/main/java/ratpack/zipkin/internal/ZipkinHttpClientImpl.java
```diff
@@ -110,23 +100,22 @@ public void close() {
...
-            // TODO cannot do handler.handleSend() as the span was created out-of-scope
-            span.name(parser.spanName(adapter, captor));
-            parser.requestTags(adapter, captor, span);
         });
     }
 
-    private void streamedResponseWithSpan(Result<StreamedResponse> response, Span span) {
+    private void streamedResponseWithSpan(Result<StreamedResponse> response, AtomicReference<Span> ref) {
+        Span span = ref.get();
+        if (span == null) return;
```
codefromthecrypt: I suspect this won't be the case unless there was an exception in the request pipeline..

src/main/java/ratpack/zipkin/ServerTracingModule.java
```diff
@@ -52,56 +52,47 @@ protected void configure() {
...
-            .serviceName(serviceName)
-            .port(port);
-    InetAddress address = configAddress != null ? configAddress : getSiteLocalAddress();
-    if (address.getAddress().length == 4) {
-      builder.ipv4(ByteBuffer.wrap(address.getAddress()).getInt());
-    } else if (address.getAddress().length == 16) {
-      builder.ipv6(address.getAddress());
+    Endpoint.Builder builder = Endpoint.builder();
+    if (!builder.parseIp(configAddress)) {
+      builder = Platform.get().localEndpoint().toBuilder();
```
llinder, line 82: nice!

src/main/java/ratpack/zipkin/internal/MethodCapturingRequestSpec.java
llinder: Ideally the client would have more callbacks for these things but I think this is the closest thing we have to an 'onRequestSent' callback.

llinder: I spent some time today trying to get the ITHttpClient tests to pass.  In the end I was able to get all tests to pass except requests with nested spans.  I even tried switching the CurrentTraceContext to the RatpackCurrentTraceContext implementation.  This also failed but in different ways.  The end result is that the current span was never found in the CurrentTraceContext so the span parentId would always be null.

Ratpack is very peculiar about what thread things are accessed on which is probably a source of the troubles.

At any rate, I will keep plugging along with this and try to figure out a solution.

llinder: I should also have mentioned, I think we need to at the very least start the parent span within an ExecHarness.yield or run method so the span context is populated from a Ratpack managed thread.  I'm not sure we can do this with the current tests.   Unfortunately when I tried that by rewriting one of the tests it still failed :(

hyleung: What's the thinking behind having this class be responsible for invoking `handleSend` instead of having it in `ZipkinHttpClientImpl`? It seems odd to have this class "reach in" to invoke a method on the handler. 

This class (`MethodCapturingRequestSpec`). I kind of think of this class as being a workaround for Ratpack's API - `RequestSpec` doesn't allow you to actually get at the HTTP method that's being used, which is why this class is required. 

Regardless of which approach is taken, it still won't be as nice as having a proper callback. Unfortunately, as far as I know Ratpack doesn't have any callbacks. I spent some time trying to get interceptors added (https://github.com/ratpack/ratpack/issues/975) but keeping the PR in sync with changes upstream was getting harder and harder and I shelved it to work on other stuff.

hyleung:
Superseded by #31

PR #31, opened 2017-05-14
Spike towards using http adapters

llinder:
continuation from https://github.com/hyleung/ratpack-zipkin/pull/30


llinder:
Down to 2 failures.  The http client redirect failure is due to the use of `wiretap` where its only called once for the final response and not for each redirect.  I will investigate if there is another option but might have to ping Ratpack maintainers on this.

The other failure is likely just something I broke and will look into later.

llinder:
@adriancole @hyleung tests are all green. Let me know what you think.  There will be some followup work when Brave 4.3.1 is releases but I'm ok refactoring things for that if you want to merge these changes now.  Or we can wait a bit, either way is fine with me.

src/main/java/ratpack/zipkin/internal/ZipkinHttpClientImpl.java
```diff
@@ -111,65 +113,165 @@ public void close() {
...
+    static final class WrappedRequestSpec implements RequestSpec {
 
-        DefaultRequestSpecNameAdapter(MethodCapturingRequestSpec request) {
-            this.request = request;
+        private final RequestSpec delegate;
+        private final ZipkinHttpClientImpl client;
+        private final AtomicReference<Span> span;
+        private HttpMethod capturedMethod;
+
+        WrappedRequestSpec(ZipkinHttpClientImpl client, RequestSpec spec,  AtomicReference<Span> span) {
```
hyleung: @llinder what do you think about just passing just what this class needs (i.e. handler and injector) instead of the whole client?

codefromthecrypt:
fyi Brave 4.3.1 is out the door and synced to central now.. this is the officially release :)

hyleung:
@adriancole sweet.

@llinder I'm ok with merging this if you feel like it's in a good spot. I wouldn't mind taking a crack at doing any refactoring that's necessary for Brave 4.3.1 - not been able to spend much time on this project lately and it might be a good way for me to re-familiarize myself with the code.

llinder:
@hyleung turned out the only change with 4.3.1 was to import HTTP specific instrumentation library.  After that everything compiled and tests pass.

What do you think about revisiting some of the tests?  Specifically it would be nice to port the HttpClient tests to use the OkHttp MockWebServer instead of mocking the client APIs.

src/main/java/ratpack/zipkin/ServerTracingModule.java
```diff
@@ -52,56 +52,47 @@ protected void configure() {
...
         .sampler(config.sampler)
         .currentTraceContext(new RatpackCurrentTraceContext())
-        .localEndpoint(buildLocalEndpoint(config.serviceName, serverConfig.getPort(), serverConfig.getAddress()))
+        .localEndpoint(buildLocalEndpoint(config.serviceName, serverConfig.getPort(),
+            serverConfig.getAddress()))
         .localServiceName(config.serviceName)
         .reporter(config.spanReporter)
         .build();
+    return HttpTracing.newBuilder(tracing)
+        .clientParser(config.clientParser)
```
codefromthecrypt, line 71: don't forget config.serverSampler!

hyleung:
@llinder @adriancole unless there's anything else to add to this PR, gonna merge it!

codefromthecrypt:
Do it!

hyleung:
Closes #30 (supersedes)

hyleung:
@llinder following up on this http client tests you mentioned - are you referring to `ZipkinHttpClientImplSpec`?

Issue #10, opened 2016-09-16
Add support for nested local spans


hyleung:
Possibly look at https://github.com/openzipkin/brave/commit/c97129e70168762665ff3ced3cfdd022536b600c for pointers.

hyleung:
@schlosna (cc @adriancole)  Had a look at the commit for`InheritableServerClientAndLocalSpanState` - thought I could model my implementation after it. Had a quick look and one thing that might make it a bit easier  could be to use something like a marker interface here (https://github.com/openzipkin/brave/commit/c97129e70168762665ff3ced3cfdd022536b600c#diff-7be6b1b0b4fbe00b0bdf7b070442c90bR93) instead of checking for an instance of a concrete class. That way, an implementer could implement `ServerClientAndLocalSpanState` and "declare" that they're capable of supporting nested local spans by just adding the marker interface. In my case, while I suppose I could extend `InheritableServerClientAndLocalSpanState`, but `ThreadLocal`s aren't really gonna work with Ratpack.

If that seems reasonable, I'm happy to open and issue/submit a PR to make it happen.

codefromthecrypt:
Mentioned this many times on different pull requests, but not on this
thread.

Inheritable... is a workaround to a design flaw in brave's context
propagation. While clever we really shouldn't add more duct tape on this.

Google agreed to share their purpose-built context, including publishing a
zero dependencies jar as of version 1.0.1

http://www.grpc.io/grpc-java/javadoc/io/grpc/Context.html

Let's just fix the root problem :)

hyleung:
@adriancole ah I see. Sorry to make you repeat yourself again! There's so much going on with Zipkin these days it's hard to keep track of things.

codefromthecrypt:
no worries. I mentioned it here, as I know it gets lost in comments :)

hyleung:
Hmm..ok so I guess grpc `Context` makes `InheritableServerClientAndLocalSpanState` go away then? Would `LocalTracer` then just assume that nested local spans are supported? i.e. this part of `LocalTracer` (https://github.com/openzipkin/brave/blob/master/brave-core/src/main/java/com/github/kristofa/brave/LocalTracer.java#L116) would just check to see if there's a current local span, if so, use that as the parent - otherwise, use the current server span.

codefromthecrypt:
In the future design, all spans are equal.

Ex there is no difference between a server parent and a non server parent.
When you create a new span, it is either a new trace or a child of what was
[...]

PR #33, opened 2017-05-18
Test for #10 deep nested local spans

llinder:


src/test/groovy/ratpack/zipkin/support/TestReporter.groovy
```diff
@@ -3,18 +3,19 @@ package ratpack.zipkin.support
 import zipkin.Span
 import zipkin.reporter.Reporter
 
+import java.util.concurrent.ConcurrentLinkedDeque
 
 class TestReporter implements Reporter<Span> {
 
-	private final List<Span> spans = new LinkedList<>()
+	private final ConcurrentLinkedDeque<Span> spans = new ConcurrentLinkedDeque<>()
```
hyleung, line 10: Is the choice here for thread-safety?

llinder: Yeah just for thread-safety.  Same pattern used by the Brave integration tests.

Right now most of the tests are single threaded with the exception of the ParallelBatch tests so probably not necessary but it feels cleaner to have it :)

PR #35, opened 2017-05-18
Fixes #28 parallel batch tracing with test

### LINKED ISSUE #28: 2.x Add ParallelBatch support
Originally submitted for v1 in #23. Was part of #24, but pulled out to be done in a separate PR.

See. 9db02da

llinder:
/cc @beckje01 @edwardsoo-ss

### DIFF
--- a/src/main/java/ratpack/zipkin/TracedParallelBatch.java
+++ b/src/main/java/ratpack/zipkin/TracedParallelBatch.java
@@ -1,0 +1,39 @@
+package ratpack.zipkin;
+
+import brave.propagation.TraceContext;
+import java.util.Arrays;
+import ratpack.exec.Promise;
+import ratpack.exec.util.ParallelBatch;
+import ratpack.zipkin.internal.RatpackCurrentTraceContext;
+
+public final class TracedParallelBatch<T> {
+
+  private final Iterable<? extends Promise<T>> promises;
+
+  private TracedParallelBatch(Iterable<? extends Promise<T>> promises) {
+    this.promises = promises;
+  }
+
+  static <T> TracedParallelBatch<T> of(Iterable<? extends Promise<T>> promises) {
+    return new TracedParallelBatch<>(promises);
+  }
+
+  static <T> TracedParallelBatch<T> of(Promise<T>... promise) {
+    return new TracedParallelBatch<>(Arrays.asList(promise));
+  }
+
+  static <T> ParallelBatch<T> of(final TraceContext context, Iterable<? extends Promise<T>> promises) {
+    return ParallelBatch.of(promises)
+        .execInit(execution ->
+            execution.add(new RatpackCurrentTraceContext.TraceContextHolder(context)));
+  }
+
+  static <T> ParallelBatch<T> of(final TraceContext context, Promise<T>... promise) {
+    return of(context, Arrays.asList(promise));
+  }
+
+  public ParallelBatch<T> withContext(final TraceContext context) {
+    return of(context, promises);
+  }
+
+}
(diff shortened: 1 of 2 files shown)


edwardsoo-ss:
:+1: :cool:

src/main/java/ratpack/zipkin/internal/RatpackCurrentTraceContext.java
```diff
@@ -23,7 +23,12 @@ public TraceContext get() {
     return registrySupplier.get()
         .maybeGet(TraceContextHolder.class)
         .map(TraceContextHolder::getContext)
-        .orElse(null);
+        .orElseGet(() -> (Execution.isManagedThread())
+          ? Execution.current()
```
hyleung: @llinder I had a question about this from the original PR from @edwardsoo-ss - what's the difference between using `Execution.current()` and `registrySupplier.get()`? `registrySupplier` is just a method ref for `Exection::current` anyway.

llinder: This does indeed seem redundant.  The tests pass without this so I'm going to remove it.  @edwardsoo-ss is there something else we are overlooking with this?

src/main/java/ratpack/zipkin/TracedParallelBatch.java
```diff
@@ -0,0 +1,39 @@
...
+  }
+
+  static <T> TracedParallelBatch<T> of(Promise<T>... promise) {
+    return new TracedParallelBatch<>(Arrays.asList(promise));
+  }
+
+  static <T> ParallelBatch<T> of(final TraceContext context, Iterable<? extends Promise<T>> promises) {
+    return ParallelBatch.of(promises)
+        .execInit(execution ->
+            execution.add(new RatpackCurrentTraceContext.TraceContextHolder(context)));
```
llinder, line 28: Thinking about this more I don't think this is quite right either.

What we need to ensure is that the context is swapped out when the promise is executed and reset afterwards... similar to https://github.com/hyleung/ratpack-zipkin/blob/master/src/main/java/ratpack/zipkin/internal/RatpackCurrentTraceContext.java#L30

llinder: Then again the docs seem to indicate that each batch receives a forked execution which presumably means its not ever shared.  If that is true then there doesn't seem to be a reason to worry about reseting a previous context.

src/main/java/ratpack/zipkin/internal/RatpackCurrentTraceContext.java
edwardsoo-ss: I checked and I agree that was redundant, we only needed to make the `TraceContextHolder` / `TypedValue` class to be public. my bad

src/main/java/ratpack/zipkin/TracedParallelBatch.java
edwardsoo-ss: that is also my understanding of ParallelBatch from experimenting: 
* the forked executions do not "inherit" registry from the caller execution
* registry change in each forked execution is not shared afterwards

hyleung:
Code change itself looks good to me (tho' can't say I've actually pulled the branch and tried it).

I'm assuming the usage won't be much different than what was described in the original PR (https://github.com/hyleung/ratpack-zipkin/pull/23#issue-219058173)?  // @llinder @edwardsoo-ss

llinder:
@hyleung I think this is good to merge.  The test coverage should give some reassurance that it works.  As for the example usage it would look more like this now:
```
TracedParallelBatch.of(promise1, promise2)
  .withContext(parent.context())
  .yieldAll()
```

So basically works just like `ParallelBatch` except users need to pass the tracing context.  I will keep a note to update the README with an example and explanation of what happens under the covers.

PR #36, opened 2017-05-23
Fix scope closing bug

hyleung:
I noticed during testing that my client spans were being "orphaned" - i.e. they didn't belong to the same trace as the server span. Digging into it a bit, I think that the server span was being closed prematurely.

Instead of using the try-with-resources block - which seems to close the
scope too early (possibly because the code inside it is async?),
creating a SpanInScope and closing it on the callback that's invoked
before the server response is sent.

Also adds Ok-Http MockServer dependency for testing.

### DIFF
--- a/src/main/java/ratpack/zipkin/ServerTracingModule.java
+++ b/src/main/java/ratpack/zipkin/ServerTracingModule.java
@@ -52,6 +52,7 @@ public class ServerTracingModule extends ConfigurableModule<ServerTracingModule.
 
 
     bind(HttpClient.class).annotatedWith(Zipkin.class).to(ZipkinHttpClientImpl.class);
+    bind(ZipkinHttpClientImpl.class);
 
     Multibinder.newSetBinder(binder(), HandlerDecorator.class).addBinding()
         .toProvider(() -> HandlerDecorator.prepend(serverTracingHandlerProvider.get()));
--- a/src/main/java/ratpack/zipkin/internal/DefaultServerTracingHandler.java
+++ b/src/main/java/ratpack/zipkin/internal/DefaultServerTracingHandler.java
@@ -33,16 +33,14 @@ public final class DefaultServerTracingHandler implements ServerTracingHandler {
   public void handle(Context ctx) throws Exception {
     Request request = ctx.getRequest();
     final Span span = handler.handleReceive(extractor, request);
-
+    final Tracer.SpanInScope scope = tracing.tracer().withSpanInScope(span);
     ctx.getResponse().beforeSend(response -> {
       handler.handleSend(response, null, span);
       span.finish();
+      scope.close();
     });
-
-    try(Tracer.SpanInScope ws = tracing.tracer().withSpanInScope(span)) {
-      span.start();
-      ctx.next();
-    }
+    span.start();
+    ctx.next();
   }
 
   static final class HttpAdapter extends brave.http.HttpServerAdapter<Request, Response> {
--- a/src/test/groovy/ratpack/zipkin/ServerTracingModuleSpec.groovy
+++ b/src/test/groovy/ratpack/zipkin/ServerTracingModuleSpec.groovy
@@ -1,6 +1,9 @@
 package ratpack.zipkin
 
 import brave.http.HttpSampler
+import okhttp3.mockwebserver.MockResponse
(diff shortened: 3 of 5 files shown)


src/main/java/ratpack/zipkin/internal/DefaultServerTracingHandler.java
```diff
@@ -33,16 +33,13 @@ public DefaultServerTracingHandler(HttpTracing httpTracing) {
...
       handler.handleSend(response, null, span);
-      span.finish();
+      scope.close();
     });
-
-    try(Tracer.SpanInScope ws = tracing.tracer().withSpanInScope(span)) {
-      span.start();
-      ctx.next();
-    }
+    span.start();
```
hyleung, line 42: As I mentioned above, think this `try-with-resources` block actually completes almost immediately.

src/test/groovy/ratpack/zipkin/ServerTracingModuleSpec.groovy
```diff
@@ -368,4 +371,49 @@ class ServerTracingModuleSpec extends Specification {
 		then:
 			assertThat(reporter.getSpans()).isNotEmpty()
 	}
+
+	def 'Should nest client span and server span should have the same trace id'() {
+		given:
+			def webServer = new MockWebServer()
+			webServer.start()
+			webServer.enqueue(new MockResponse().setResponseCode(200))
+			def url = webServer.url("/")
```
hyleung, line 380: @llinder I do want to eventually use the same approach in the http client tests, as you suggested. Didn't do it in this PR because I haven't figured out a nice way to get Guice to give me an instance of `ZipkinHttpClientImpl`. Here, because it's in an embedded app, it's pretty straightforward.

src/main/java/ratpack/zipkin/ServerTracingModule.java
```diff
@@ -52,6 +52,7 @@ protected void configure() {
 
 
     bind(HttpClient.class).annotatedWith(Zipkin.class).to(ZipkinHttpClientImpl.class);
+    bind(ZipkinHttpClientImpl.class);
```
hyleung, line 55: Needed so that I could just as Guice or an instance of this in tests.

hyleung:
Last call for comments on this one - I'll give it another day and then merge it. cc. @llinder

src/main/java/ratpack/zipkin/internal/DefaultServerTracingHandler.java
```diff
@@ -33,16 +33,13 @@ public DefaultServerTracingHandler(HttpTracing httpTracing) {
   public void handle(Context ctx) throws Exception {
     Request request = ctx.getRequest();
     final Span span = handler.handleReceive(extractor, request);
-
+    final Tracer.SpanInScope scope = tracing.tracer().withSpanInScope(span);
     ctx.getResponse().beforeSend(response -> {
       handler.handleSend(response, null, span);
-      span.finish();
+      scope.close();
```
llinder: I think we still need `span.finish()` to record timings and such.

src/main/java/ratpack/zipkin/internal/DefaultServerTracingHandler.java
llinder: Good catch.

src/test/groovy/ratpack/zipkin/ServerTracingModuleSpec.groovy
llinder: For the client tests I would imagine we would just instantiate ZipkinHttpClientImpl manually like I did for the brave integration tests.  For example:
https://github.com/hyleung/ratpack-zipkin/blob/master/src/test/java/brave/http/ITTracingFeature_Client.java#L39

I'm not advocating that we should tackle this now though :)  The fixes you have here are more important to have merged.

llinder:
Only think that is missing is the `span.finish()` call.  Other than that 👍

src/test/groovy/ratpack/zipkin/ServerTracingModuleSpec.groovy
hyleung: @llinder ah ic...forgot that you'd done it like that. I'll refactor the client tests in a separate PR.

PR #40, opened 2017-09-15
Upgrade to Span V2

llinder:
Upgrades things to the Span V2 model.

Also started pushing the trace ID into the MDC so its easily available for log correlation.

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -1,4 +1,4 @@
 version=2.1.1-SNAPSHOT
 
-braveVersion=4.7.1
+braveVersion=4.9.1
 ratpackVersion=1.4.6
--- a/gradle.properties
+++ b/gradle.properties
@@ -1,4 +1,4 @@
 version=2.1.1-SNAPSHOT
 
-braveVersion=4.5.0
+braveVersion=4.7.1
 ratpackVersion=1.4.6
--- a/gradle.properties
+++ b/gradle.properties
@@ -1,4 +1,4 @@
 version=2.1.1-SNAPSHOT
 
-braveVersion=4.9.1
+braveVersion=4.9.2
 ratpackVersion=1.4.6
--- a/src/main/java/ratpack/zipkin/ServerTracingModule.java
+++ b/src/main/java/ratpack/zipkin/ServerTracingModule.java
@@ -147,16 +147,21 @@ public class ServerTracingModule extends ConfigurableModule<ServerTracingModule.
      *
      * If not set, defaults to {@link Reporter#NOOP}.
      *
-     * @param reporter the reporter
+     * @param reporter a V2 reporter
      *
      * @return the config
      */
-    public Config spanReporter(final Reporter<Span> reporter) {
+    public Config spanReporterV2(final Reporter<Span> reporter) {
       this.spanReporter = reporter;
       return this;
     }
 
-    /** @deprecated please use {@link #spanReporter(Reporter)} */
+    /** @deprecated please use {@link #spanReporterV2(Reporter)}
+     *
+     * @param reporter a V1 reporter
+     *
+     * @return the config
+     * */
     // Until this is removed, we need a mandatory dep on io.zipkin.reporter:zipkin-reporter
     // due to overloading requiring access to all overloaded types
     @Deprecated
@@ -267,8 +272,8 @@ public class ServerTracingModule extends ConfigurableModule<ServerTracingModule.
      *
      * Defaults to {@link B3Propagation.Factory}.
      *
-     * @param propagationFactory
-     * @return
+     * @param propagationFactory the Zipkin propagation factory
+     * @return the config
      */
     public Config propagationFactory(final Propagation.Factory propagationFactory) {
       this.propagationFactory = propagationFactory;
--- a/src/main/java/ratpack/zipkin/ServerTracingModule.java
+++ b/src/main/java/ratpack/zipkin/ServerTracingModule.java
@@ -24,11 +24,14 @@ import brave.http.HttpSampler;
 import brave.http.HttpServerParser;
 import brave.http.HttpTracing;
 import brave.internal.Platform;
+import brave.propagation.B3Propagation;
+import brave.propagation.Propagation;
 import brave.sampler.Sampler;
 import com.google.inject.Provider;
 import com.google.inject.Provides;
 import com.google.inject.Singleton;
 import com.google.inject.multibindings.Multibinder;
+import java.net.InetAddress;
 import ratpack.api.Nullable;
 import ratpack.guice.ConfigurableModule;
 import ratpack.handling.HandlerDecorator;
@@ -39,9 +42,7 @@ import ratpack.zipkin.internal.RatpackCurrentTraceContext;
 import ratpack.zipkin.internal.ZipkinHttpClientImpl;
 import zipkin2.Endpoint;
 import zipkin2.Span;
-import zipkin.reporter.Reporter;
-
-import java.net.InetAddress;
+import zipkin2.reporter.Reporter;
 
 /**
  * Module for Zipkin distributed tracing.
@@ -92,6 +93,7 @@ public class ServerTracingModule extends ConfigurableModule<ServerTracingModule.
                                  serverConfig.getAddress()))
                              .localServiceName(config.serviceName)
                              .spanReporter(config.spanReporter)
+                             .propagationFactory(config.propagationFactory)
                              .build();
     return HttpTracing.newBuilder(tracing)
                       .clientParser(config.clientParser)
@@ -119,7 +121,7 @@ public class ServerTracingModule extends ConfigurableModule<ServerTracingModule.
    */
   public static class Config {
     private String serviceName = "unknown";
-    private Reporter<Span> spanReporter = (Span s) -> {};
+    private Reporter<Span> spanReporter = Reporter.NOOP;
     private Sampler sampler = Sampler.NEVER_SAMPLE;
     private HttpSampler serverSampler = HttpSampler.TRACE_ID;
     private HttpSampler clientSampler = HttpSampler.TRACE_ID;
@@ -127,6 +129,7 @@ public class ServerTracingModule extends ConfigurableModule<ServerTracingModule.
     private HttpClientParser clientParser = new HttpClientParser();
     private HttpServerParser serverParser = new HttpServerParser();
     private SpanNameProvider spanNameProvider = (req, pathBinding) -> req.getMethod().getName();
+    private Propagation.Factory propagationFactory = B3Propagation.FACTORY;
 
     /**
      * Set the service name.
@@ -150,6 +153,27 @@ public class ServerTracingModule extends ConfigurableModule<ServerTracingModule.
      */
     public Config spanReporter(final Reporter<Span> reporter) {
       this.spanReporter = reporter;
+      return this;
+    }
+
+    /** @deprecated please use {@link #spanReporter(Reporter)} */
+    // Until this is removed, we need a mandatory dep on io.zipkin.reporter:zipkin-reporter
+    // due to overloading requiring access to all overloaded types
+    @Deprecated
+    public Config spanReporter(final zipkin.reporter.Reporter<zipkin.Span> reporter) {
+      if (reporter == zipkin.reporter.Reporter.NOOP) {
+        this.spanReporter = Reporter.NOOP;
+        return this;
+      }
+      this.spanReporter = new Reporter<zipkin2.Span>() {
+        @Override public void report(zipkin2.Span span) {
+          reporter.report(zipkin.internal.V2SpanConverter.toSpan(span));
+        }
+
+        @Override public String toString() {
+          return reporter.toString();
+        }
+      };
       return this;
     }
 
@@ -235,6 +259,19 @@ public class ServerTracingModule extends ConfigurableModule<ServerTracingModule.
      */
     public Config spanNameProvider(final SpanNameProvider spanNameProvider) {
       this.spanNameProvider = spanNameProvider;
+      return this;
+    }
+
+    /**
+     * Set the {@link Propagation.Factory}.
+     *
+     * Defaults to {@link B3Propagation.Factory}.
+     *
+     * @param propagationFactory
+     * @return
+     */
+    public Config propagationFactory(final Propagation.Factory propagationFactory) {
+      this.propagationFactory = propagationFactory;
       return this;
     }
   }
--- a/src/main/java/ratpack/zipkin/ServerTracingModule.java
+++ b/src/main/java/ratpack/zipkin/ServerTracingModule.java
@@ -37,8 +37,8 @@ import ratpack.server.ServerConfig;
 import ratpack.zipkin.internal.DefaultServerTracingHandler;
 import ratpack.zipkin.internal.RatpackCurrentTraceContext;
 import ratpack.zipkin.internal.ZipkinHttpClientImpl;
-import zipkin.Endpoint;
-import zipkin.Span;
+import zipkin2.Endpoint;
+import zipkin2.Span;
 import zipkin.reporter.Reporter;
 
 import java.net.InetAddress;
@@ -91,7 +91,7 @@ public class ServerTracingModule extends ConfigurableModule<ServerTracingModule.
                              .localEndpoint(buildLocalEndpoint(config.serviceName, serverConfig.getPort(),
                                  serverConfig.getAddress()))
                              .localServiceName(config.serviceName)
-                             .reporter(config.spanReporter)
+                             .spanReporter(config.spanReporter)
                              .build();
     return HttpTracing.newBuilder(tracing)
                       .clientParser(config.clientParser)
@@ -107,7 +107,7 @@ public class ServerTracingModule extends ConfigurableModule<ServerTracingModule.
   }
 
   private static Endpoint buildLocalEndpoint(String serviceName, int port, @Nullable InetAddress configAddress) {
-    Endpoint.Builder builder = Endpoint.builder();
+    Endpoint.Builder builder = Endpoint.newBuilder();
     if (!builder.parseIp(configAddress)) {
       builder = Platform.get().localEndpoint().toBuilder();
     }
@@ -119,7 +119,7 @@ public class ServerTracingModule extends ConfigurableModule<ServerTracingModule.
    */
   public static class Config {
     private String serviceName = "unknown";
-    private Reporter<Span> spanReporter = Reporter.NOOP;
+    private Reporter<Span> spanReporter = (Span s) -> {};
     private Sampler sampler = Sampler.NEVER_SAMPLE;
     private HttpSampler serverSampler = HttpSampler.TRACE_ID;
     private HttpSampler clientSampler = HttpSampler.TRACE_ID;
--- a/src/test/groovy/ratpack/zipkin/ServerTracingModuleSpec.groovy
+++ b/src/test/groovy/ratpack/zipkin/ServerTracingModuleSpec.groovy
@@ -1,7 +1,6 @@
 package ratpack.zipkin
 
 import brave.SpanCustomizer
-import brave.Tracing
 import brave.http.HttpSampler
 import okhttp3.mockwebserver.MockResponse
 import okhttp3.mockwebserver.MockWebServer
@@ -62,7 +61,7 @@ class ServerTracingModuleSpec extends Specification {
 								.sampler(Sampler.create(1f))
 								.clientSampler(HttpSampler.TRACE_ID)
 								.serverSampler(HttpSampler.TRACE_ID)
-								.spanReporter(Reporter.NOOP)
+								.spanReporterV2(Reporter.NOOP)
 					})
 				}).handlers {
 					chain ->
@@ -83,7 +82,7 @@ class ServerTracingModuleSpec extends Specification {
 						config
 								.serviceName("embedded")
 								.sampler(Sampler.create(1f))
-								.spanReporter(zipkin.reporter.Reporter.NOOP)
+								.spanReporterV2(zipkin.reporter.Reporter.NOOP)
 					})
 				}).handlers {
 					chain ->
@@ -141,7 +140,7 @@ class ServerTracingModuleSpec extends Specification {
 						config
 								.serviceName("embedded")
 								.sampler(Sampler.create(1f))
-								.spanReporter(reporter)
+								.spanReporterV2(reporter)
 					})
 				}).handlers {
 					chain ->
@@ -168,7 +167,7 @@ class ServerTracingModuleSpec extends Specification {
 						config
 								.serviceName("embedded")
 								.sampler(Sampler.create(1f))
-								.spanReporter(reporter)
+								.spanReporterV2(reporter)
 					})
 				}).handlers {
 					chain ->
@@ -206,7 +205,7 @@ class ServerTracingModuleSpec extends Specification {
 						config
 								.serviceName("embedded")
 								.sampler(Sampler.create(1f))
-								.spanReporter(reporter)
+								.spanReporterV2(reporter)
 					})
 				}).handlers {
 					chain ->
@@ -247,7 +246,7 @@ class ServerTracingModuleSpec extends Specification {
 						config
 								.serviceName("embedded")
 								.sampler(Sampler.create(1f))
-								.spanReporter(reporter)
+								.spanReporterV2(reporter)
 					})
 				}).handlers {
 					chain ->
@@ -276,7 +275,7 @@ class ServerTracingModuleSpec extends Specification {
 						config
 								.serviceName("embedded")
 								.sampler(Sampler.create(1f))
-								.spanReporter(reporter)
+								.spanReporterV2(reporter)
 					})
 				}).handlers {
 					chain ->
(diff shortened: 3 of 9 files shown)


src/main/java/ratpack/zipkin/internal/RatpackCurrentTraceContext.java
```diff
@@ -21,26 +25,47 @@ public RatpackCurrentTraceContext() {
...
+    } else {
+      registrySupplier.get().add(ContextHolder.EMPTY);
+      return () -> {};
+    }
   }
 
-  public static final class TraceContextHolder {
+  private static final class ContextHolder {
+
+    private static final ContextHolder EMPTY = new ContextHolder(null);
```
llinder: An empty value to avoid extra object creation every time the last Scope is closed.

src/main/java/ratpack/zipkin/internal/RatpackCurrentTraceContext.java
```diff
@@ -21,26 +25,47 @@ public RatpackCurrentTraceContext() {
...
+      final Optional<ContextHolder> previous = registrySupplier.get()
+          .maybeGet(ContextHolder.class);
+
+      registrySupplier.get().add(new ContextHolder(currentSpan));
+      MDC.put(MCD_TRACEID_KEY, currentSpan.traceIdString());
+
+      return () -> {
+        if (previous.isPresent() && previous.get().context != null) {
+          registrySupplier.get().add(previous);
+          MDC.put(MCD_TRACEID_KEY, previous.get().context.traceIdString());
```
llinder: We could probably make the MDC stuff optional if others feel its not always desirable.  I thought I would just throw it in here and get feedback first before doing extra work to make it optional :)

src/test/groovy/ratpack/zipkin/ServerTracingModuleSpec.groovy
```diff
@@ -55,7 +53,7 @@ class ServerTracingModuleSpec extends Specification {
 								.sampler(Sampler.create(1f))
 								.clientSampler(HttpSampler.TRACE_ID)
 								.serverSampler(HttpSampler.TRACE_ID)
-								.spanReporter(Reporter.NOOP)
+								.spanReporter({Span s -> }) // noop
```
llinder: Reporter.NOOP uses Span V1.  Not sure if there is an equivalent type for Span V2 but I couldn't find one.

src/main/java/ratpack/zipkin/TracedParallelBatch.java
```diff
@@ -60,8 +60,7 @@ private TracedParallelBatch(Iterable<? extends Promise<T>> promises) {
    */
   public static <T> ParallelBatch<T> of(final TraceContext context, Iterable<? extends Promise<T>> promises) {
     return ParallelBatch.of(promises)
-        .execInit(execution ->
-            execution.add(new RatpackCurrentTraceContext.TraceContextHolder(context)));
+        .execInit(execution -> execution.add(RatpackCurrentTraceContext.wrap(context)));
```
llinder: Really wish this wasn't referring to `RatpackCurrentTraceContext` directly here.  It basically makes it impossible to use a different implementation.

Long term I think this API probably needs some deeper thought.  For now I'm just punting on it but should probably make a ticket and try to come up with a better solution.

hyleung: @llinder by "different implementation", do you mean a different implementation of `CurrentTraceContext`?

src/main/java/ratpack/zipkin/internal/RatpackCurrentTraceContext.java
hyleung: @llinder personally, I'm a fan of MDC, so I'm not opposed to having it in. But if it's not too tied into the upgrade to V2 spans, you could just split it out into a separate PR and we can work on getting this one merged before tackling that.

src/main/java/ratpack/zipkin/TracedParallelBatch.java
llinder: Yeah, meant different implementation of CurrentTraceContext.  Right now this is tightly coupled to RatpackCurrentTraceContext which in hind sight feels less than ideal. :(

src/main/java/ratpack/zipkin/internal/RatpackCurrentTraceContext.java
llinder: Sure I will make it a different PR.  Actually that part is what got me started on this because I need something like that sooner than the V2 stuff.  Figured I would try and tackle the V2 stuff as long as I was in the code though.

src/test/groovy/ratpack/zipkin/ServerTracingModuleSpec.groovy
codefromthecrypt: there isn't at the moment, we could add NOOP2? :)

src/test/groovy/ratpack/zipkin/ServerTracingModuleSpec.groovy
```diff
@@ -265,8 +261,8 @@ class ServerTracingModuleSpec extends Specification {
 		then:
 			reporter.getSpans().size() == 1
 			Span span = reporter.getSpans().get(0)
-			span.binaryAnnotations.find { it -> it.key == TraceKeys.HTTP_STATUS_CODE } != null
-			span.binaryAnnotations.find { it -> it.key == Constants.ERROR } != null
+			span.tags().find { it -> it.key == TraceKeys.HTTP_STATUS_CODE } != null
```
codefromthecrypt, line 328: might be a map assertion?

src/test/groovy/ratpack/zipkin/internal/RatpackCurrentTraceContextSpec.groovy
```diff
@@ -61,4 +62,23 @@ class RatpackCurrentTraceContextSpec extends Specification {
             traceContext == null
     }
 
+    def 'When TraceContext is null the context should be cleared'() {
```
codefromthecrypt: thx

codefromthecrypt (approved):
LG!

llinder:
I've moved all of the CurrentTraceContext changes including the MDC parts into this PR https://github.com/hyleung/ratpack-zipkin/pull/41

I'm going to revise the Span V2 PR to build on that but only include V2 stuff.

src/main/java/ratpack/zipkin/TracedParallelBatch.java
hyleung: @llinder what use case can you think of where we'd require a different implementation of `CurrentTraceConext`? In any case, as long as this stuff is an implementation detail of (in this case `TraceParallelBatch` and it doesn't leak out, I figure we're free to change it later if we feel it's necessary.

src/test/groovy/ratpack/zipkin/ServerTracingModuleSpec.groovy
llinder: Thanks for the suggestion!  Map assertions are awesome... I clearly don't use assertj enough :)

src/main/java/ratpack/zipkin/TracedParallelBatch.java
llinder: An example would have been if I needed a custom CurrentTraceContext that does some magic for non Ratpack things.  One case is the MDC addition but there are others such as some cross execution stuff between Ratpack and Cassandra.  In order to solve the Ratpack Execution -> Cassandra Execution hops I'm likely going to have to modify the CurrentTraceContext.  Honestly not sure how yet but I know its been a problem for us so I will be looking into it more soon.

In general though, the CurrentTraceContext is pluggable in Brave, so maintaining the "pluggability" in the Ratpack instrumentation would be ideal.

llinder:
Ok I've updated this PR and build on top of https://github.com/hyleung/ratpack-zipkin/pull/41.  Once 41 is merged I will rebase this PR to remove the MDC commit.

hyleung:
@llinder I'm gonna try to do a little testing on this PR. Might take me a little while because I haven't been keeping up with Zipkin developments so need to catch up on the v2 stuff - but I haven't forgotten about this!

gradle.properties
```diff
@@ -1,4 +1,4 @@
 version=2.0.2-SNAPSHOT
 
-braveVersion=4.5.0
+braveVersion=4.7.1
```
codefromthecrypt: 4.8.0 :)

src/main/java/ratpack/zipkin/ServerTracingModule.java
```diff
@@ -33,8 +33,8 @@
 import ratpack.zipkin.internal.DefaultServerTracingHandler;
 import ratpack.zipkin.internal.RatpackCurrentTraceContext;
 import ratpack.zipkin.internal.ZipkinHttpClientImpl;
-import zipkin.Endpoint;
-import zipkin.Span;
+import zipkin2.Endpoint;
+import zipkin2.Span;
 import zipkin.reporter.Reporter;
```
codefromthecrypt: zipkin2.reporter.Reporter

src/main/java/ratpack/zipkin/ServerTracingModule.java
```diff
@@ -88,7 +88,7 @@ private static Endpoint buildLocalEndpoint(String serviceName, int port, @Nullab
    */
   public static class Config {
     private String serviceName = "unknown";
-    private Reporter<Span> spanReporter = Reporter.NOOP;
+    private Reporter<Span> spanReporter = (Span s) -> {};
```
codefromthecrypt: Reporter.NOOP is here now

src/test/groovy/ratpack/zipkin/ServerTracingModuleSpec.groovy
```diff
@@ -55,7 +53,7 @@ class ServerTracingModuleSpec extends Specification {
 								.sampler(Sampler.create(1f))
 								.clientSampler(HttpSampler.TRACE_ID)
 								.serverSampler(HttpSampler.TRACE_ID)
-								.spanReporter(Reporter.NOOP)
+								.spanReporter({Span s -> })
```
codefromthecrypt: zipkin2.reporter.Reporter.NOOP is back

src/test/groovy/ratpack/zipkin/support/TestReporter.groovy
```diff
@@ -1,6 +1,6 @@
 package ratpack.zipkin.support
 
-import zipkin.Span
+import zipkin2.Span
 import zipkin.reporter.Reporter
```
codefromthecrypt: zipkin2.reporter.Reporter

codefromthecrypt (commented):
made some notes now that we have a v2 native reporter (you can drop any dep on io.zipkin.reporter:zipkin-reporter for io.zipkin.reporter2:zipkin-reporter)

src/main/java/ratpack/zipkin/ServerTracingModule.java
hyleung: [`Config#spanReporter`](https://github.com/llinder/ratpack-zipkin/blob/19a6724734d5a5f71967bb3f3795cff6f57bc9bd/src/main/java/ratpack/zipkin/ServerTracingModule.java#L118-L121)  needs to be updated to return `zipkin2.reporter.Reporter` too. cc/ @llinder 

[Edit] Well, I guess it wouldn't need to be updated since the class name would be the same.

hyleung: I suppose this would mean that v1 `Reporter`s aren't going to be usable anymore, which would make this a breaking change.

codefromthecrypt:
Well you can also support reportet v1 api (via tracing.builder.reporter) or
v1 format (via encoder)
https://github.com/openzipkin/brave/blob/master/brave/README.md#zipkin-v1-setup

codefromthecrypt:
Basically for compat I would recommend accepting a v1 reporter but
preferring v2 internally. You can convert a v1 reporter to a v2 one (ex
look at Tracing.Builder.reporter for a hint)

codefromthecrypt:
ok I made some changes here which seem to work. feel free to
copy/paste instead of trying to merge my commit if it helps
https://github.com/llinder/ratpack-zipkin/pull/2

llinder:
I've been struggling to keep up on my own PR sorry :(  I this covers the feedback and also incorporates @adriancole changes.  Looks like the code coverages is dropping so I will try to see what we are missing.

BTW I spent some time trying to get the ITHttpServer integration tests working.  It looks like I might need to do some refactoring on this to make it work with Ratpack.  Once I get a window of time again I will try to pick it up again and make  PR upstream when I get things to work locally.

hyleung:
@llinder no need to be sorry! From the [coverage report](https://codecov.io/gh/hyleung/ratpack-zipkin/src/6b40c4aa34735856f1699b59e99fd69d286d8243/src/main/java/ratpack/zipkin/ServerTracingModule.java#L122...140), it looks like it's a couple of missed code paths in the deprecated `spanReporter` method on the Config. The no-op case would probably the only test you could argue would be valuable. If we cover that case and the coverage check still fails, I'm cool with just letting it go.

codefromthecrypt:
thanks lance. you're a champ!

llinder:
All green!  I changed the legacy reporter handling a bit to push this up to the tracer itself.  The result should be the same, just slightly changed how the Tracing instance is built.

src/main/java/ratpack/zipkin/ServerTracingModule.java
```diff
@@ -59,14 +59,21 @@ protected void configure() {
...
         .currentTraceContext(new RatpackCurrentTraceContext())
         .localEndpoint(buildLocalEndpoint(config.serviceName, serverConfig.getPort(),
             serverConfig.getAddress()))
-        .localServiceName(config.serviceName)
-        .reporter(config.spanReporter)
-        .build();
+        .localServiceName(config.serviceName);
+
+    // for backwards comparability, prefer legacy reporter if defined otherwise use the
+    if (config.legacyReporter != null) {
```
codefromthecrypt: Ps i dont remember how I coded before but I think I probably coded to avoid a check like this. This check requires zipkin 1 in the runtime classpath (vs just compile). It isnt obvious until after you try to remove the dep ;) 300 KiB might not matter to people and easy enough to rewire later if asked.

codefromthecrypt (approved):
Good work!

hyleung:
@llinder I pulled the branch to do a quick test - pretty much just dropped the updated dependency in and everything worked w/o any other code changes, so in terms of legacy compatibility, we're good. Still need to get myself setup to test the v2 stuff.

src/main/java/ratpack/zipkin/ServerTracingModule.java
hyleung: huh, so if we removed zipkin 1 from the runtime class path, do we get an exception or something?

codefromthecrypt:
yeah you'd get a classnotfound. In brave, we guard with an internal api
Platform.zipkin1Present I think. however, if the reporter is converted
prior to this step, then there's no chance of classnotfound

hyleung:
@adriancole wow, i wouldn't have expected that by just looking at the code! I tried it just now and I somehow got a compile time error ¯\_(ツ)_/¯ . Anyhow, I'm not too bothered either way (if we leave the dependency in or out) - thoughts, @llinder?

codefromthecrypt:
>
> @adriancole <https://github.com/adriancole> wow, i wouldn't have expected
> that by just looking at the code! I tried it just now and I somehow got a
> compile time error ¯_(ツ)_/¯ . Anyhow, I'm not too bothered either way (if
> we leave the dependency in or out) - thoughts, @llinder
> <https://github.com/llinder>?
>
yeah basically it is splitting hairs a bit. ratpack-zipkin does need a
compile dep, but a user configuring ratpack-zipkin could go without it at
runtime, if the legacy span reporter wasn't required to be checked. Ex I
use a separate project to verify this

https://github.com/openzipkin/brave/tree/master/brave/src/it/nozipkin1

llinder:
Didn't even consider compile vs runtime dependency on Zipkin 1.  Definitely some merit in not making the dependency required.  I will revert this change and get code coverage up a different way :)

hyleung:
@llinder this'll need a rebase after #45. Are you still looking trying to make the v1 dependency not-required?

llinder:
@hyleung I added the code back that Adrian suggested so that should remove the hard v1 dependency https://github.com/hyleung/ratpack-zipkin/pull/40/files#diff-6d11129c36716a94c98d2e75a65b5a5cR154.  I've also upgrade Gradle wrapper to 2.12 so we can declare the v1 reporter as `compileOnly` so this shouldn't show up in the transitive dependencies now.

Also rebased against master so this should be good to go.

codefromthecrypt:
ps I will write release notes later, but propagationFactory is a big part of brave 4.9 going out now. Notably, this allows you to use amazon (single-header) propagation or even push other things

https://github.com/openzipkin/brave/tree/master/propagation/aws#brave-propagation-aws
https://github.com/openzipkin/brave/blob/master/brave/README.md#propagating-extra-fields

llinder:
Thanks for the heads up on 4.9.  I will look at updating this and adding the propagationFactory to the module config later today.

llinder:
Rebased and added `propagationFactory` support.  @hyleung what do you think about merge this?  Any additional gaps we should fill?

hyleung:
@llinder I'll take a quick look (hopefully) tonight - should be fine to merge unless there were a lot of changes required to add `propagationFactory`.

hyleung:
@llinder I pulled the branch and did a quick check to make sure that things work as far as backward compatibility goes. Still want to test the v2 spans but I'll need to futz about with my test app and setup a little before I can verify that. Is this something that's urgent for you or can I take a couple of days to do a few more checks?

hyleung:
Not sure if I just missed something really obvious, but I moved my test app over to `zipkin2` dependencies, but needed to add a dep with `zipkin.reporter.Reporter` for it to compile.

llinder:
@hyleung no this isn't urgent so please don't feel rushed.

As far as I know `zipkin.reporter.Reporter` dependency should only be required if your actually using the V1 reporter.  This was one of the changes I made a while ago, which removed the hard dependency on this.  If your test app uses that then it makes sense you had to add that for it to compile.  Most users going forward won't need this so it is only an optional dependency now.

hyleung:
@llinder ok, cool. As far as I know I'm not using the v1 reporter. I'll keep digging - might have to trouble you for a quick second set of 👀s. 

I haven't been keeping up to date with Zipkin developments, so I'm not quite sure how to go about getting set up to test v2 (I've been using the docker-compose stack).

codefromthecrypt:
Server exposes both endpoints so you are all good. Hopefully you have a
compile dep only building the project itself. End consumers should be able
to leave out the v1 dep but that is something to verify.

hyleung:
@adriancole hmm...i'm running the latest `docker-zipkin` but `curl localhost:9411/api/v2/services` gives me a `404`. I exec'd into the container so I'm assuming there's not some weird networking thing going on.

codefromthecrypt:
did you docker-compose pull? hit the /info endpoint to verify the version
is indeed latest?

hyleung:
@adriancole yeah, I did do a pull.

```
$  http "http://localhost:9411/info"
HTTP/1.1 200
Content-Type: application/vnd.spring-boot.actuator.v1+json;charset=UTF-8
Date: Wed, 08 Nov 2017 00:03:46 GMT
Transfer-Encoding: chunked
X-Application-Context: zipkin-server:shared:9411

{
    "zipkin": {
        "version": "2.2.1"
    }
}

$  http "http://localhost:9411/api/v2/services"
HTTP/1.1 404
Content-Length: 0
Date: Wed, 08 Nov 2017 00:04:27 GMT
X-Application-Context: zipkin-server:shared:9411
```

hyleung:
Here's the [branch](https://github.com/hyleung/ratpack-zipkin-example/tree/upgrade-to-v2) in my test project where I've been trying to upgrade to v2. When I *don't* include the v1 reporter dependency, I get a compilation error:

```
./gradlew clean check
:clean
:compileJava
/Users/hyleung/Projects/ratpack-zipkin-example/src/main/java/com/example/server/App.java:29: error: cannot access Reporter
                  .spanReporter(spanReporter());
                  ^
  class file for zipkin.reporter.Reporter not found
1 error
:compileJava FAILED
```

codefromthecrypt:
Thx will look into it and release a patch if it isnt a mutual eyes crossing
problem!

hyleung:
lol "mutual eyes crossing" - definitely stealing that!

Addition info, I'm running Docker for Mac:

Version 17.09.0-ce-mac34 (19605)
Channel: edge
c076099f52

I've also tried `docker-machine` with the same result. I've exec'd into the container so I'm assuming it isn't some weird networking issue. Looking at the logs as far as I can tell, the v2 handlers are being picked up, etc.

codefromthecrypt:
hmm I'm using curl and get this...

curl -v 192.168.99.100:9411/api/v2/services

*   Trying 192.168.99.100...

* Connected to 192.168.99.100 (127.0.0.1) port 9411 (#0)

> GET /api/v2/services HTTP/1.1

> Host: 192.168.99.100:9411

> User-Agent: curl/7.43.0

> Accept: */*

>

< HTTP/1.1 200

< X-Application-Context: zipkin-server:shared:9411

< Content-Type: application/json;charset=UTF-8

< Transfer-Encoding: chunked

< Vary: Accept-Encoding

< Date: Wed, 08 Nov 2017 06:51:02 GMT

<

* Connection #0 to host 192.168.99.100 left intact

[]

codefromthecrypt:
also just tried httpie

http 192.168.99.100:9411/api/v2/services

HTTP/1.1 200

Content-Encoding: gzip

Content-Type: application/json;charset=UTF-8

Date: Wed, 08 Nov 2017 06:54:50 GMT

Transfer-Encoding: chunked

Vary: Accept-Encoding

X-Application-Context: zipkin-server:shared:9411


[]

hyleung:
Hmm.... @llinder if you were able to verify that the v2 stuff works as expected (i.e. you're able to verify that traces are being recorded properly), I'm fine with that (I'll figure out what's up with my environment later). 

I'm still not able to get my demo project compiling w/o adding v1 dependencies (https://github.com/hyleung/ratpack-zipkin-example/tree/upgrade-to-v2) - if you've got the time, I'd really appreciate it if you could take a look to see if I'm just missing something really obvious.

codefromthecrypt:
there was a v1 encoding bug, can you update to brave 4.9.1 (which depends on zipkin2 2.2.2 which has the fix)? Sorry about that. was missing a test and backfilled a bunch

codefromthecrypt:
(ps this is unrelated to the compiling issue... sorry no coffee yet)

hyleung:
@adriancole This PR already upgrades to Brave 4.9.1 so think we're ok. I've also updated my `docker-zipkin` so that it's running 2.2.2. For whatever reason I'm still getting that 404 on `/services` but the traces seem to be recording fine (at least with OkHttp) ¯\_(ツ)_/¯.

Re. the compilation issue, still need to include `io.zipkin.reporter:zipkin-reporter:1.1.2` tho'.

codefromthecrypt:
can you give me the cli input to build in such a way I can repro the sample
problem?

I did ./gradlew clean build and it is angry about "Cannot perform signing
task ':signArchives' because it has no configured signatory"

hyleung:
@adriancole argh 🤦‍♂️ I'm sorry - totally forgot about that, Ive been meaning to fix that. Need to update `build.gradle` so that it doesn't try to sign local builds. My bad 😞

llinder:
@hyleung I think I see the issue with the compile errors when the `zipkin-reporter:1.1.2` dependency isn't included.  It appears that we need to change the method name for `spanReporter` to something else so the compiler can disambiguate the argument types.  Simply changing it to `spanReporterV2` does the trick but is a bit ugly.  If you have a suggestion on the naming for this let me know.  For now I will just update my PR to use `spanReporterV2` for lack of a better name.

llinder:
@hyleung I've tested the name change to verify that the `zipkin-reporter:1.1.2` dependency is no longer required.  I've also tested the `ratpack-zipkin-example` project against a local running Zipkin 2.3.0 server and spans are reported correctly.

image

I'm fairly confident that things are working as expected now.  If you feel we should check anything else let me know and I will jump on it as quick as I can.

hyleung:
@llinder thanks! Looks good to me. Think we can merge it now :)

I'll update the README separately before doing a release (assuming we're in no particular hurry).

Re. the `spanReporterV2` naming, I think it's probably fine. And actually I was wondering how long we would be supporting both v1 and v2 reporters. Until it becomes really painful to support both, I figure it's probably more user-friendly to keep supporting both for a while.  We could even add a `spanReporterV1`, which can supersede the deprecated `spanReporter` so that at least we're being very explicit about which version is expected.

codefromthecrypt:
Agree on method name problem. In brave reporter was v1 so we made v2
spanReporter for this reason. Drat should have noticed!

On 13 Nov 2017 4:41 am, "Ho Yan Leung" <notifications@github.com> wrote:

Merged #40 <https://github.com/hyleung/ratpack-zipkin/pull/40>.

—
You are receiving this because you were mentioned.
Reply to this email directly, view it on GitHub
<https://github.com/hyleung/ratpack-zipkin/pull/40#event-1337747830>, or mute
the thread

PR #41, opened 2017-09-15
Add current trace ID to the MDC

llinder:
- this adds the current trace ID to the MDC and removes it when the scope is closed.
- adds singleton empty context holder to avoid extra object creation when the context is null.
- marks access to the TraceContextHolder as deprecated as we want to move away from leaking internal details of the RatpackCurrentTraceContext.

### DIFF
--- a/src/main/java/ratpack/zipkin/internal/RatpackCurrentTraceContext.java
+++ b/src/main/java/ratpack/zipkin/internal/RatpackCurrentTraceContext.java
@@ -3,10 +3,13 @@ package ratpack.zipkin.internal;
 import brave.propagation.CurrentTraceContext;
 import brave.propagation.TraceContext;
 import java.util.function.Supplier;
+import org.slf4j.MDC;
 import ratpack.exec.Execution;
 import ratpack.registry.MutableRegistry;
 
 public final class RatpackCurrentTraceContext extends CurrentTraceContext {
+
+  private static final String TRACE_ID_KEY = "TraceId";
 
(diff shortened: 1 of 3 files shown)


beckje01 (approved):

llinder:
@hyleung if this looks good to you would you mind cutting a new release for 2.0.1 with just the MDC change?  Then we can follow up with the Span V2 PR and make a separate release for that when its ready.  Probably want to make the V2 changes a 2.1 release but we can discuss that  in the other PR.

hyleung:
@llinder yup! I'll do a quick sanity check on the branch tonight and get this PR merged and released.

src/main/java/ratpack/zipkin/internal/RatpackCurrentTraceContext.java
```diff
@@ -22,25 +25,56 @@ public RatpackCurrentTraceContext() {
...
 
-  public static final class TraceContextHolder {
+  /**
+   * Used by TracedParallelBatch where its used to wrap a TraceContext and puts it in the
+   * registry for the forked execution.  This is marked deprecated as we prefer not to
+   * expose details of the RatpackCurrentTraceContext implementation.
+   *
+   * @param traceContext
+   * @return
+   */
```
hyleung, line 63: @llinder ok, I think I kinda see why you're pointing this out as being less than ideal. Guess we'll need to revisit `TracedParallelBatch` at some point. There's a couple of javadoc warnings at build time but I will take care of that in another commit.

Issue #26, opened 2017-04-26
2.x - Allow customization of span naming logic for traces


hyleung:
In v1, there was a `SpanNameProvider` interface that a user could provide an implementation of via the module config.

```
public interface SpanNameProvider {
    String spanName(HttpRequest request);
}
```

e.g.

```
config
.serviceName("ratpack-demo")
.spanNameProvider(new SpanNameProvider() {
        @Override
         public String spanName(final HttpRequest request) {
                      return "foo" + request.getHttpMethod();
         }
})
```

For v2, we should provide something similar.  `SpanNameProvider` still exists but we've removed the ability to configure it. Also it still has a lot of the feel of Brave http (adapters, etc.) - so now might be a good opportunity to redesign that stuff (thinking maybe a `@FunctionalInterface`).

codefromthecrypt:
I currently have a plugin with 3 methods spanName requestTags and
responseTags which default but can be overridden as a unit.

This allows you to use the same plugin to coerce any http library (similar
to brave 3) what is different is that the parsing is lazy and also the
input can have other data.

Ex currently thinking brave.http.HttpRequestContext which wraps calls ex
req.method() would call the underlying. However you could subtype this.. ex
RatpackHttpRequestContext which could possibly also expose a connection
which could aid building more sophisticated tags. It could also expose the
raw request object.

Moreover the ResponseContext could carry an error message or throwable.
This allows someone to write the "error" tag differently across any library
defaulting to something sensible.

Will ping with concrete code which will be soon like today

Im currently jigging this across other libraries and want to make sure it
works.

codefromthecrypt:
made some notes here. After looking both ways I ditched the wrapping
approach which I tried for a few hours today. I think this one is best, but
let me know. Regardless, we can choose to make span name parser an atom or
a part of a composite plugin.

https://github.com/openzipkin/brave/pull/350/files#r113442950

hyleung:
@adriancole sorry, I didn't get a chance to look at this until today - the link above just goes to the PR (maybe because of a rebase?). Anyhow - is the idea that we'd extend `brave.ClientHandler` or `brave.Serverhandler`? Or would it be `brave.http.HttpAdapter`?

llinder:
This can be customized with Brave 4.3.1 by providing a custom HttpParser and override the `spanName` method.  The custom HttpParser can be set through the ServerTracingModule `serverParser` and `clientParser`.

[...]

Issue #27, opened 2017-04-26
2.x - Add support for customizing tags on request/response


hyleung:
This is another feature that was in v1, but removed as part of #24. Users used to be able to inject their own logic for adding `KeyValueAnnotations` via module configuration. E.g.

```
config
    .serviceName("ratpack-demo")
    .requestAnnotations(request ->
        Lists.newArrayList(KeyValueAnnotation.create("uri", request.getUri()))
    )
    .responseAnnotations(response ->
        Lists.newArrayList(KeyValueAnnotation.create("foo", "bar"))
    );
```

[Update] Going to look at customizing tags first...

codefromthecrypt:
Fyi i have work in progress calling these requestTags and responseTags.
Will ping on details but essentially resurrecting brave-http

llinder:
With Brave 4.3.1 custom annotations can be provided through a custom `HttpParser` and overriding the `request(adapter, req, customizer)` and `response(adapter, res, error, customizer)` methods.

In these methods the `customizer` parameter is of type `SpanCustomizer` which allows users to add tags and annotations to the span.

Users can provide a custom `HttpParser` through the `ServerTracingModule` `clientParser` and `serverParser` configuration properties.

@hyleung do you think this sufficiently covers the ability to customize the span?

hyleung:
@llinder yeah I think for now it's probably good enough. Or at least, we can defer until we get a better idea of how it "feels". Where I wanted to go with this was something more "typical" for Ratpack - i.e. using a lambda to customize the behavior.

hyleung:
@adriancole are there any updates on the request/response Tag stuff?

> Fyi i have work in progress calling these requestTags and responseTags.
Will ping on details but essentially resurrecting brave-http

codefromthecrypt:
> @adriancole <https://github.com/adriancole> are there any updates on the
> request/response Tag stuff?
>
> Fyi i have work in progress calling these requestTags and responseTags.
> Will ping on details but essentially resurrecting brave-http
>
>
https://github.com/openzipkin/brave/tree/master/instrumentation/http#span-data-policy
is latest and greatest

hyleung:
Revisiting this after #52 - thinking if I do add something like this, it would be more straightforward to just have a function (or functions) that take a `SpanCustomizer` and let folks do whatever they like.

PR #42, opened 2017-09-20
Add support for span name customization.

### LINKED ISSUE #39: Pass Context to HttpAdapter instead of Request/Response
There is a desire to access parts of the context to customize annotations in the [HttpParser](https://github.com/openzipkin/brave/blob/master/instrumentation/http/src/main/java/brave/http/HttpParser.java).  Rather than pulling the request and the response off the context to send to [HttpAdapter](https://github.com/hyleung/ratpack-zipkin/blob/master/src/main/java/ratpack/zipkin/internal/DefaultServerTracingHandler.java#L46) we could instead send the Context so its available to the HttpParser.

pseudo code
```
static final class HttpAdapter extends brave.http.HttpServerAdapter<Context, Context> {

    @Override public String method(Context context) {
      return context.getRequest().getMethod().getName();
    }

    @Override public String path(Context context) {
      return context.getRequest().getPath();
    }

    @Override public String url(Context context) {
      return context.getRequest().getUri();
    }

    @Override public String requestHeader(Context context, String name) {
      return context.getRequest().getHeaders().get(name);
    }

    @Override public Integer statusCode(Context context) {
      return context.getResponse().getStatus().getCode();
    }
  }
```

hyleung:
Add support for span name customization via module configuration. Module
configuration now can take a function that returns a custom span name
using information from the request and response.

Since the stock Ratpack response doesn't provide access to the
PathBinding (that's only accessible from the Context), we're using
wrapper to expose that information. If this ends up being supported
upstream, we can eventually get rid of this hack.

Closes #39 and #26


src/main/java/ratpack/zipkin/internal/DefaultServerTracingHandler.java
```diff
@@ -7,62 +7,103 @@
...
+      handler.handleSend(wrappedResponse, null, span);
       span.finish();
       scope.close();
     });
     span.start();
     ctx.next();
   }
 
-  static final class HttpAdapter extends brave.http.HttpServerAdapter<Request, Response> {
+  private static class ServerRequestImpl implements ServerRequest {
```
hyleung, line 60: This diff is totally messed up, but the adapter has been moved out. Just felt a little tidier.

src/main/java/ratpack/zipkin/ServerRequest.java
```diff
@@ -0,0 +1,36 @@
+package ratpack.zipkin;
+
+import ratpack.http.Headers;
+import ratpack.http.HttpMethod;
+
+/**
+ * Read-only API for wrappers around Ratpack's {@link ratpack.http.Request}.
+ */
+public interface ServerRequest {
```
hyleung, line 9: Strictly speaking, this isn't necessary because Ratpack's `Request` gives us access to everything we would need. But I've kept this around because it felt like a read-only API in the lambda sort of communicates better what you can/cannot do (i.e don't try to mutate things in the lambda). Also for some reason it felt weird to have the lambda take a Ratpack `Request` for one param and a `ServerResponse` for the other.

src/main/java/ratpack/zipkin/ServerTracingModule.java
```diff
@@ -92,8 +96,10 @@ private static Endpoint buildLocalEndpoint(String serviceName, int port, @Nullab
     private Sampler sampler = Sampler.NEVER_SAMPLE;
     private HttpSampler serverSampler = HttpSampler.TRACE_ID;
     private HttpSampler clientSampler = HttpSampler.TRACE_ID;
+
     private HttpClientParser clientParser = new HttpClientParser();
     private HttpServerParser serverParser = new HttpServerParser();
+    private SpanNameProvider spanNameProvider = (req,res) -> req.getMethod().getName();
```
hyleung: Default Span name is just the method name.

hyleung:
@llinder if you have any thoughts on this one, I'd greatly appreciate the feedback!

llinder:
Nice!  I read through this and nothing really suck out to comment on.  This should be useful for several folks that had some hacks on the old version to get access to the path binding.  Thanks for working on this!

src/main/java/ratpack/zipkin/SpanNameProvider.java
```diff
@@ -0,0 +1,15 @@
...
+ */
+@FunctionalInterface
+public interface SpanNameProvider {
+  /**
+   * Given a request and response, return the name to be used for the Span.
+   * @param request the request
+   * @param response the response
+   * @return the Span name
+   */
+  String spanName(ServerRequest request, ServerResponse response);
```
codefromthecrypt: response is odd to have here. seems the reason is to expose the path binding.. if so, make a comment or doc it so folks don't forget

hyleung: @adriancole hmmm...yeah I see your point. TBH, there's probably not a hell of a lot of value in having `ServerResponse` here (I think I just have it this way because it seemed more "symmetrical"). I could just put the `PathBinding` in there - at least then it's pretty clear that that's the *only* reason that parameter exists (i.e. don't expect to be able to get anything else from the response).

codefromthecrypt:
The other thing is sometimes a failure happens before the response. Naming
based on the request makes sure the span is named even when a response
never comes.

hyleung:
@adriancole ah, of course - I hadn't thought about that. That being the case, what do you think about making the second param `Optional<PathBinding>`?

llinder:
As @adriancole pointed out, applying the span name in the `beforeSend` handler is less than ideal.   I just don't see a way around this if we still want access to the `PathBinding`.  On the other hand, if `beforeSend` is never called its likely a framework bug so the chances of it not being called seems rather low.   Making it an `Optional<PathBinding>` does seem like a good idea though.  If I remember correctly, if Ratpack doesn't find a route that matches the request, the `PathBinding` might be null.   We should probably add some tests for this case if we don't have them yet.

If folks are concerned about naming the span in `beforeSend` it might be worth making the `SpanNameProvider` configurable so the `RatpackServerTracingHandler` can make a decision to sample in the request vs the response.

codefromthecrypt:
one thing I should mention is that it isn't bad to "rename" a span later,
just that it should have a name before the request goes out. For example,
nothing's stopping doing span.name("fooo") with a better name at response
time..  just if it has no name at all there could be more issues.

PS sampling needs to happen before the request goes out, as otherwise
downstream won't see the decision. (sampling is by default trace-scoped not
span-scoped though we can do span.abandon to throw away spans in a trace we
don't like)

hyleung:
Ah...how about calling that span name provider (with `Optional<PathBinding>`) *twice* to set the span name - once when the request is received (with `Optional.empty()`) and once beforeSend (with a `PathBinding` value). That way we ensure that there'll be a span name regardless of whether there's a  failure.

The function that provides the span name would look something like this...

```
SpanNameProvider spanNameProvider = (req, pathBindingOpt) -> pathBindingOpt
    .map(pathBinding -> pathBinding.getDescription())
    .orElse(req.getPath());
```

hyleung:
Alright, I reworked this to pass an `Optional<PathBinding>` //cc @llinder

codefromthecrypt:
For users trying to be efficient, they can check if the latter is present
and skip vs compute twice overriding. That or we can overload to do the
same. Or do nothing :)

src/main/java/ratpack/zipkin/internal/DefaultServerTracingHandler.java
```diff
@@ -46,11 +46,11 @@ public DefaultServerTracingHandler(final HttpTracing httpTracing, final SpanName
   public void handle(Context ctx) throws Exception {
     ServerRequest request = new ServerRequestImpl(ctx.getRequest());
     final Span span = handler.handleReceive(extractor, request);
+    span.name(spanNameProvider.spanName(request, Optional.empty()));
     final Tracer.SpanInScope scope = tracing.tracer().withSpanInScope(span);
 
     ctx.getResponse().beforeSend(response -> {
-      String name = spanNameProvider.spanName(request, Optional.of(ctx.getPathBinding()));
-      span.name(name);
+      span.name(spanNameProvider.spanName(request, Optional.of(ctx.getPathBinding())));
```
llinder: Not sure if `ctx.getPathBinding()` can ever be null but we can guard against an NPE here if we use `Optional.ofNullable`.

hyleung: Btw, I couldn't figure out a way to force the failure case (where there's a failure before the response is sent) in a unit test ¯\_(ツ)_/¯

llinder:
Looks good!  Thanks again for the work on this.

hyleung:
@llinder 👍 just did a quick (manual) test to make sure it's working as expected (https://github.com/hyleung/ratpack-zipkin-example/commit/0d426dcaaa34cfdba0b4502a5af10d1469457b2b). Did occur to me that #27 should probably provide access to the `PathBinding`, too. 

I'll let this sit for another day and then do a release and maybe start on that.

PR #45, opened 2017-10-15
Make Guice components singletons

llinder:
Currently the Guice components are not singletons.  Besides the unnecessary instantiation, this also breaks things like the CurrentSpanCustomizer pattern where each CurrentSpanCustomizer was receiving a new tracer.  Since each tracer has its own recorder, additional tags are always added to a new span instead of the current span.

In addition to adding the singleton scopes this also exposes `Tracing`, `Tracer` and `SpanCustomizer` on the guice registry for convenience.

### DIFF
--- a/src/main/java/ratpack/zipkin/ServerTracingModule.java
+++ b/src/main/java/ratpack/zipkin/ServerTracingModule.java
@@ -15,6 +15,9 @@
  */
 package ratpack.zipkin;
 
+import brave.CurrentSpanCustomizer;
+import brave.SpanCustomizer;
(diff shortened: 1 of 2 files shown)


llinder:
@hyleung if this looks ok would you mind merging and making a 2.0.2 release?

hyleung (approved):

llinder:
@hyleung  thanks a ton for reviewing this so quick!

hyleung:
@llinder no problem at all! I think 2.0.2 should be up by now...

Issue #46, opened 2017-10-23
Action being executed multiple times in HttpClient.request


ptong-samsung:
https://github.com/hyleung/ratpack-zipkin/blob/master/src/main/java/ratpack/zipkin/internal/ZipkinHttpClientImpl.java#L85

Action is executed twice, once in action.append, and once in wrapped request.
If the user is using functions like requestSpec.headers.add, duplicate headers will be added to the request. This causes error in some of the server. For example, AWS ELB will return 400 when it founds multiple Authroization header.

hyleung:
@ptong-samsung thanks for reporting this - do you have any workarounds available to you that can mitigate the issue? As you can probably tell from looking at that code, there was a fair bit of yak shaving required to get client tracing working so there's a chance we can't yak-shave our way out of this one. Ratpack *has* added support for client request/response interceptors and at some point we were going to revisit our client tracing implementation in this library to make use of that (#44).

ptong-samsung:
@hyleung Thanks. Yes we just changing everything to use headers.set so its fine. Just want to create this to keep trace of the issue.

llinder:
I see whats happening here.  I think I can fix this quick.

llinder:
For our use case it would be good to get this fixed before Ratpack 1.6 with interceptors since we still have some code bases one 1.4.5 and updating takes a bit of time.

hyleung:
@llinder I've merged and done a release that includes your fix (#47) - it's still propagating out to Maven central etc, the version is 2.1.0. I'm a little behind on day-job stuff, but when I get a minute, I'll write up some quick release notes (really it's just #47 and #42/#27)

cc/ @ptong-samsung

PR #47, opened 2017-10-26
Remove double action execution on client requests

llinder:
Resolves issue #46 

cc/ @rbiga @ptong-samsung

### DIFF
--- a/src/main/java/ratpack/zipkin/internal/ZipkinHttpClientImpl.java
+++ b/src/main/java/ratpack/zipkin/internal/ZipkinHttpClientImpl.java
@@ -43,8 +43,8 @@ import ratpack.http.client.StreamedResponse;
 public final class ZipkinHttpClientImpl implements HttpClient {
 
     private final HttpClient delegate;
-    final HttpClientHandler<WrappedRequestSpec, Integer> handler;
-    final TraceContext.Injector<MutableHeaders> injector;
+    private final HttpClientHandler<WrappedRequestSpec, Integer> handler;
+    private final TraceContext.Injector<MutableHeaders> injector;
 
     @Inject
     public ZipkinHttpClientImpl(final HttpClient delegate, final HttpTracing httpTracing) {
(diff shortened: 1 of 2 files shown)


src/test/groovy/ratpack/zipkin/internal/ZipkinHttpClientImplSpec.groovy
```diff
@@ -87,8 +88,30 @@ class ZipkinHttpClientImplSpec extends Specification {
...
+			}
+
+		then:
+			reporter.spans.size() == 1
+			Span span = reporter.spans.get(0)
+			span.name == "get"
+
+		then:
+			RecordedRequest request = webServer.takeRequest()
+			request.headers.values("X-TEST").size() == 1
```
llinder, line 112: Before the change in `ZipkinHttpClientImpl` in this PR it would return 2 headers for `X-TEST`.

src/main/java/ratpack/zipkin/internal/ZipkinHttpClientImpl.java
```diff
@@ -80,26 +80,25 @@ public void close() {
...
 
     @Override
     public Promise<StreamedResponse> requestStream(URI uri, Action<? super RequestSpec> action) {
-        AtomicReference<Span> span = new AtomicReference<>();
+        final AtomicReference<Span> span = new AtomicReference<>();
+
         return delegate
-            .requestStream(uri, action.append(requestSpec -> {
-                WrappedRequestSpec captor = new WrappedRequestSpec(this.handler, this.injector, requestSpec, span);
+            .requestStream(uri, (RequestSpec requestSpec) -> {
```
jeff-blaisdell, line 95: Looking at this now this makes more sense to me.
  
For some reason before I had it in my head that I shouldn't modify the original action, and I needed to add mine to execute in addition; believing the dual execution needed to happen.

src/test/groovy/ratpack/zipkin/internal/ZipkinHttpClientImplSpec.groovy
jeff-blaisdell: Like the test

jeff-blaisdell:
👍 great change

codefromthecrypt (approved):
Nice

hyleung (approved):
Thanks for fixing this, @llinder!

PR #48, opened 2017-10-28
Add support for configuring tag customizers.

### LINKED ISSUE #27: 2.x - Add support for customizing tags on request/response

hyleung:
Closes #27 

Adds support for configuring tag customizers. 

Looks something like this in use...
```
config
    .requestTagCustomizer(request -> Collections.singletonList(Pair.of("foo", "bar")))
    .responseTagCustomizer((response, pathBinding) ->
                           pathBinding
                           .getTokens()
                           .entrySet()
                           .stream()
                           .map(entry -> Pair.of(entry.getKey(), entry.getValue()))
                           .collect(Collectors.toList()));
```

Some of this tag customization is already possible using `HttpServerParser`. My thinking for adding this was to make it a little more convenient and more RatPack-y (i.e. specifying a lambda instead of needing to define an anonymous class). Additionally, on the response-side of things, you can access the `PathBinding`, which (currently) you can't using `HttpServerParser` (that's only because work to make that happen hasn't been done yet).

The main use case I had in my mind (other than to make it a bit more RatPack-y) was to be able to add the `PathBinding` tokens to the tags. So, for example, you could query for traces with certain parameter values.

### DIFF
--- a/src/main/java/ratpack/zipkin/ResponseTagCustomizer.java
+++ b/src/main/java/ratpack/zipkin/ResponseTagCustomizer.java
@@ -1,0 +1,18 @@
+package ratpack.zipkin;
+
+import ratpack.func.Pair;
+import ratpack.http.Response;
+import ratpack.path.PathBinding;
+
+import java.util.List;
+
+@FunctionalInterface
+public interface ResponseTagCustomizer {
+  /**
+   * Return the tags for a response and pathbinding.
+   * @param response the response
+   * @param pathBinding the path binding
+   * @return a list of key/value pairs to add to the span as tags.
+   */
+  List<Pair<String, String>> tags(Response response, PathBinding pathBinding);
+}
--- a/src/main/java/ratpack/zipkin/internal/DefaultServerTracingHandler.java
+++ b/src/main/java/ratpack/zipkin/internal/DefaultServerTracingHandler.java
@@ -10,17 +10,22 @@ import javax.inject.Inject;
 
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
+import ratpack.func.Pair;
 import ratpack.handling.Context;
 import ratpack.handling.Handler;
 import ratpack.http.Headers;
 import ratpack.http.HttpMethod;
 import ratpack.http.Request;
 import ratpack.http.Response;
+import ratpack.zipkin.RequestTagCustomizer;
+import ratpack.zipkin.ResponseTagCustomizer;
 import ratpack.zipkin.ServerRequest;
 import ratpack.zipkin.ServerTracingHandler;
 import ratpack.zipkin.SpanNameProvider;
 
+import java.util.List;
 import java.util.Optional;
+import java.util.function.BiFunction;
 
 /**
  * {@link Handler} for Zipkin tracing.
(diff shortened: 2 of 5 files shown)


src/test/groovy/ratpack/zipkin/ServerTracingModuleSpec.groovy
```diff
@@ -492,4 +496,79 @@ class ServerTracingModuleSpec extends Specification {
...
+                        })
+                }
+            }
+		when:
+            app.test { t ->
+                t.get("/say/hello")
+            }
+		then:
+            assertThat(reporter.getSpans()).isNotEmpty()
+            def span = reporter.getSpans().first()
```
codefromthecrypt, line 623: Ps switching your test span reporter to use v2 format will make this easier to read.

codefromthecrypt (commented):
Maybe add a concrete use case as the PR description and possibly type docs? For example, what is a tag policy you can now achieve but couldnt before. (Ex lookup by X)

src/test/groovy/ratpack/zipkin/ServerTracingModuleSpec.groovy
hyleung: @adriancole I guess that would be something that depends on the upgrade to v2 spans (#40)?

codefromthecrypt:
>
> @adriancole <https://github.com/adriancole> I guess that would be
> something that depends on the upgrade to v2 spans (#40
> <https://github.com/hyleung/ratpack-zipkin/pull/40>)?
>
right.. forgot that wasn't merged

src/test/groovy/ratpack/zipkin/ServerTracingModuleSpec.groovy
hyleung: @adriancole I rebased after #40 was merged and re-worked the tests a bit. Also tried to flesh out the PR description a bit.

hyleung:
Gonna let this sit for the weekend. If there aren't any objections by then, I'm gonna merge and do a release!

src/main/java/ratpack/zipkin/ResponseTagCustomizer.java
```diff
@@ -0,0 +1,18 @@
...
+
+@FunctionalInterface
+public interface ResponseTagCustomizer {
+  /**
+   * Return the tags for a response and pathbinding.
+   * @param response the response
+   * @param pathBinding the path binding
+   * @return a list of key/value pairs to add to the span as tags.
+   */
+  List<Pair<String, String>> tags(Response response, PathBinding pathBinding);
```
codefromthecrypt, line 17: here's a couple things to consider, which may not be important enough, but whatev :)

from a microperf pov, accepting a SpanCustomizer or otherwise is better than returning a list except in the case of empty list. even in empty list, it is no difference unless you create an iterator. for example, list is nice as you can use  `for (int i.. ` syntax to scan the list which doesn't allocate an iterator and benchmarks faster.

The design of passing a `SpanCustomizer` as a final arg allows you to skip iteration completely. Ex you just push things in that you want.

Ex. below will bench much faster..
```
if (foo) customizer.tag("foo", foo);
```
than this
```
result = new ArrayList();
if (foo) result.add(entry("foo", foo));
return result;
}

... somewhere else..
tags = parser.tags(...)
for (entry tag: tags) span.tag(entry._1, entry_.2)
```

src/main/java/ratpack/zipkin/internal/DefaultServerTracingHandler.java
```diff
@@ -31,24 +36,36 @@
...
   }
 
   @Override
   public void handle(Context ctx) throws Exception {
     ServerRequest request = new ServerRequestImpl(ctx.getRequest());
-    final Span span = handler.handleReceive(extractor, request);
-    span.name(spanNameProvider.spanName(request, Optional.empty()));
+    final Span span = handler.handleReceive(extractor, request)
+                       .name(spanNameProvider.spanName(request, Optional.empty()));
+    for( Pair<String, String> kv: requestTagCustomizer.tags(request)) {
```
codefromthecrypt, line 60: handleReceive currently dodges parsing unless it can't. It would be neat to find a way to integrate what you have into the normal parser so that it can be controlled similarly. Otherwise you can consider guarding as parsing tags can sometimes add overhead you don't want in noop case.

ex. (if span.isNoop) // don't parse tags

codefromthecrypt (commented):
made some perf-related notes.. one thing to consider is that these parsers (even the name one I think) aren't using any of the existing parsing infra. They perform overlapping functionality which can lead to double-tagging if someone is using the stock brave parsers. It would be nice if the customizers here could become an implementation detail of a standard customizer. That allows more flexibility in the future, for example filtering tags globally based on http policy.

codefromthecrypt:
so basic challenge is if you can make a `HttpServerParser` that is actually composed of this.

ex internally save these functions as implementation details of a `HttpServerParser`. the user would hopefully know no difference.

hyleung:
@adriancole thanks so much for taking the time to provide so much extra context (I really appreciate it!). I think there's probably some Brave concepts that I didn't fully grasp. Is the intent of `HttpServerParser` to allow users to express their span customization "policy" in a way that just "plugs into" Brave?

codefromthecrypt:
Yep. Whether users are coding themselves or are using something that does
it for them. The original req/resp objects are available so that you can
use the same class even if you have to occasionally cast into an apm type

https://github.com/openzipkin/brave/blob/master/instrumentation/http/README.md#span-data-policy


On 18 Nov 2017 12:43, "Ho Yan Leung" <notifications@github.com> wrote:

@adriancole <https://github.com/adriancole> thanks so much for taking the
time to provide so much extra context (I really appreciate it!). I think
there's probably some Brave concepts that I didn't fully grasp. Is the
intent of HttpServerParser to allow users to express their span
customization "policy" in a way that just "plugs into" Brave?

—
You are receiving this because you were mentioned.

Reply to this email directly, view it on GitHub
<https://github.com/hyleung/ratpack-zipkin/pull/48#issuecomment-345420431>,
or mute the thread

hyleung:
@adriancole ah ic I get it now. With this in mind, then, I'm going to put this on hold for a bit. I'll try to re-work things to make them more oriented around `HttpServerParser` first.

PR #50, opened 2017-11-12
Span v2 misc tweaks

hyleung:
Updates the README with some info about how to use Zipkin V2 (might flesh it out a bit more later).

~Also playing with the idea of leaving in `spanReporterV1` (i.e. with the goal of retaining Zipkin v1 support for longer).~

### DIFF
--- a/src/main/java/ratpack/zipkin/ServerTracingModule.java
+++ b/src/main/java/ratpack/zipkin/ServerTracingModule.java
@@ -156,14 +156,13 @@ public class ServerTracingModule extends ConfigurableModule<ServerTracingModule.
       return this;
     }
 
-    /** @deprecated please use {@link #spanReporterV2(Reporter)}
+
+    /** @deprecated please use {@link #spanReporterV2(Reporter)}}
      *
      * @param reporter a V1 reporter
      *
      * @return the config
      * */
-    // Until this is removed, we need a mandatory dep on io.zipkin.reporter:zipkin-reporter
-    // due to overloading requiring access to all overloaded types
     @Deprecated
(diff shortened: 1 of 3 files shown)


src/main/java/ratpack/zipkin/ServerTracingModule.java
```diff
@@ -156,16 +156,14 @@ public Config spanReporterV2(final Reporter<Span> reporter) {
...
      * @param reporter a V1 reporter
      *
      * @return the config
-     * */
-    // Until this is removed, we need a mandatory dep on io.zipkin.reporter:zipkin-reporter
-    // due to overloading requiring access to all overloaded types
-    @Deprecated
-    public Config spanReporter(final zipkin.reporter.Reporter<zipkin.Span> reporter) {
+     */
+    public Config spanReporterV1(final zipkin.reporter.Reporter<zipkin.Span> reporter) {
```
hyleung: @llinder any thoughts on leaving v1 support in for longer? As I mentioned in #40 before merging, I think it might be more user-friendly to leave v1 support in as long as possible...

codefromthecrypt:
Maybe instead have docs on how to write v1 format with the v2 reporter? V1
code will not have any maintenance. If someone screams we can always add it
back worst case.

hyleung:
@adriancole ah, ic...if v1 code isn't going to be maintained then yeah, I guess there's no point in continuing have the option for using a v1 reporter.

codefromthecrypt:
yeah I'd consider it just a temporary thing. So it could be deprecated for
a while to allow people to move over in case they have custom stuff. the
apis are near identical, so that period could be short

PR #52, opened 2017-12-13
Refactors tracing handler to use HttpServerParser

### LINKED ISSUE #51: Use HttpServerParser to do span name customization
The current implementation in `DefaultServerTracingHandler` customizes the span name by setting it directly into the Span. We should rework this implementation so that it uses Brave's `HttpServerParser` to implement span naming (and tagging) policy - see https://github.com/openzipkin/brave/blob/master/instrumentation/http/README.md#span-data-policy

See thread from https://github.com/hyleung/ratpack-zipkin/pull/48 for additional context

hyleung:
This commit introduces an implementaton of `HttpServerParser` and
refactors the server tracing handler such that span name customization
is done via the `HttpServerParser` instead of directly in the tracing
handler.

In order to make this work, needed to add a wrapper around the standard
Ratpack `Response` so that things like the `PathBinding` and original
`ServerRequest` are available to the `HttpServerParser`.

Closes #51

### DIFF
--- a/src/main/java/ratpack/zipkin/internal/DefaultServerTracingHandler.java
+++ b/src/main/java/ratpack/zipkin/internal/DefaultServerTracingHandler.java
@@ -44,15 +44,15 @@ public final class DefaultServerTracingHandler implements ServerTracingHandler {
   public void handle(Context ctx) throws Exception {
     ServerRequest request = new ServerRequestImpl(ctx.getRequest());
     final Span span = handler.handleReceive(extractor, request);
-    final Tracer.SpanInScope scope = tracing.tracer().withSpanInScope(span);
 
+    //place the Span in scope so that downstream code (e.g. Ratpack handlers
+    //further on in the chain) can see the Span.
+    final Tracer.SpanInScope scope = tracing.tracer().withSpanInScope(span);
     ctx.getResponse().beforeSend(response -> {
       ServerResponse serverResponse = new ServerResponseImpl(response, request, ctx.getPathBinding());
       handler.handleSend(serverResponse, null, span);
-      span.finish();
       scope.close();
     });
-    span.start();
     ctx.next();
   }
 
--- a/src/main/java/ratpack/zipkin/internal/DefaultServerTracingHandler.java
+++ b/src/main/java/ratpack/zipkin/internal/DefaultServerTracingHandler.java
@@ -16,9 +16,11 @@ import ratpack.http.Headers;
 import ratpack.http.HttpMethod;
 import ratpack.http.Request;
 import ratpack.http.Response;
+import ratpack.http.Status;
+import ratpack.path.PathBinding;
 import ratpack.zipkin.ServerRequest;
+import ratpack.zipkin.ServerResponse;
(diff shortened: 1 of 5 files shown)


codefromthecrypt:
Looks good. FYI I will be raising something soon for "http.template", which is a tag that is uniform across templated requests, though not guaranteed to be a URI template.

hyleung:
@adriancole thanks for taking a look! I'm wondering if I can get rid of all the explicit Span start/finish stuff that's in `DefaultServerTracingHandler`. As far as I can tell, that should all be taken care of by the `HttpServerHandler`. 

https://github.com/hyleung/ratpack-zipkin/blob/ae1a8367e0deddbcc5df9181e039937b097d7886/src/main/java/ratpack/zipkin/internal/DefaultServerTracingHandler.java#L46-L55

src/main/java/ratpack/zipkin/internal/DefaultServerTracingHandler.java
```diff
@@ -28,32 +30,29 @@
...
   @Override
   public void handle(Context ctx) throws Exception {
     ServerRequest request = new ServerRequestImpl(ctx.getRequest());
     final Span span = handler.handleReceive(extractor, request);
-    span.name(spanNameProvider.spanName(request, Optional.empty()));
-    final Tracer.SpanInScope scope = tracing.tracer().withSpanInScope(span);
 
+    //create a scope so that we can @Inject a SpanCustomizer into
+    //handlers and add tags, annotations, etc.
+    final Tracer.SpanInScope scope = tracing.tracer().withSpanInScope(span);
```
hyleung: I don't actually know whether this is "a thing" - i.e. wanting to be able `@Inject` a `SpanCustomizer` into a handler and add tags, etc. I guess it might be useful if it's something that you don't want to implement as a general "policy" using something like a custom `HttpServerParser` ¯\\_(ツ)_/¯

codefromthecrypt: maybe harder hitting comment could be.. "place the scan in scope so that downstream code, such as logging can see the ids"

src/main/java/ratpack/zipkin/internal/DefaultServerTracingHandler.java
```diff
@@ -28,32 +30,29 @@
...
-    span.name(spanNameProvider.spanName(request, Optional.empty()));
-    final Tracer.SpanInScope scope = tracing.tracer().withSpanInScope(span);
 
+    //place the Span in scope so that downstream code (e.g. Ratpack handlers
+    //further on in the chain) can see the Span.
+    final Tracer.SpanInScope scope = tracing.tracer().withSpanInScope(span);
     ctx.getResponse().beforeSend(response -> {
-      span.name(spanNameProvider.spanName(request, Optional.ofNullable(ctx.getPathBinding())));
-      handler.handleSend(response, null, span);
-      span.finish();
```
codefromthecrypt, line 53: I think you asked.. yes, removing explicit start/finish was a good choice!

PR #53, opened 2017-12-14
Updates to latest brave

codefromthecrypt:
much faster under load


hyleung:
@adriancole thanks! Just saw the new release last night (really good write up about the issue, too).

PR #54, opened 2017-12-19
Implements ITHttpServer test suite

### LINKED ISSUE #43: Implement support for the stock Brave http server test suite
Ref: https://github.com/hyleung/ratpack-zipkin/issues/39#issuecomment-329670462

Similar work has already been done for supporting the stock http _client_ test suite.

hyleung:
Closes #43.

Adds `ITHttpServer` test from brave http instrumentation library.

- fixes `ServerAdapter.getUrl` implementation (wasn't returning the
expected value)
- `ServerAdapter` overrides `HttpServerAdapter`'s `parseClientAddress` so that
we get the right remote endpoint

### DIFF
--- a/src/main/java/ratpack/zipkin/ServerTracingModule.java
+++ b/src/main/java/ratpack/zipkin/ServerTracingModule.java
@@ -31,7 +31,6 @@ import com.google.inject.Provider;
 import com.google.inject.Provides;
 import com.google.inject.Singleton;
 import com.google.inject.multibindings.Multibinder;
-import java.net.InetAddress;
 import ratpack.api.Nullable;
 import ratpack.guice.ConfigurableModule;
 import ratpack.handling.HandlerDecorator;
@@ -44,6 +43,7 @@ import ratpack.zipkin.internal.ZipkinHttpClientImpl;
 import zipkin2.Endpoint;
 import zipkin2.Span;
 import zipkin2.reporter.Reporter;
+import java.net.InetAddress;
 
 /**
  * Module for Zipkin distributed tracing.
@@ -276,5 +276,4 @@ public class ServerTracingModule extends ConfigurableModule<ServerTracingModule.
       return this;
     }
   }
-
 }
--- a/src/main/java/ratpack/zipkin/internal/DefaultServerTracingHandler.java
+++ b/src/main/java/ratpack/zipkin/internal/DefaultServerTracingHandler.java
@@ -8,6 +8,7 @@ import brave.http.HttpTracing;
 import brave.propagation.TraceContext;
 import javax.inject.Inject;
 
+import com.google.common.net.HostAndPort;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import ratpack.handling.Context;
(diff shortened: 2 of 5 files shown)


src/main/java/ratpack/zipkin/internal/DefaultServerTracingHandler.java
```diff
@@ -82,6 +84,20 @@ public String getPath() {
...
       return request.getHeaders();
     }
+
+    @Override
+    public String getUrl() {
+      PublicAddress publicAddress = request.get(Context.class).get(PublicAddress.class);
+      return publicAddress.builder()
+                          .path(request.getPath())
+                          .params(request.getQueryParams())
+                          .build().toString();
```
hyleung, line 94: I'm assuming that the URL we want is whatever the client "sees". [`PublicAddress`](https://ratpack.io/manual/current/api/ratpack/server/PublicAddress.html) is the Ratpack class that is used to create absolute URLs and redirects, so figured this would be appropriate.

codefromthecrypt: yep

src/main/java/ratpack/zipkin/ServerTracingModule.java
```diff
@@ -87,6 +87,10 @@ public Tracing getTracing(final HttpTracing httpTracing) {
 
   @Provides @Singleton
   public HttpTracing getHttpTracing(final Config config, final ServerConfig serverConfig) {
```
codefromthecrypt: wonder if we can't use module override instead? https://stackoverflow.com/questions/483087/overriding-binding-in-guice

codefromthecrypt (commented):
glad to see this!

hyleung:
@adriancole thanks for the hint about Guice module overrides - totally forgot that you could do that!

src/main/java/ratpack/zipkin/internal/ServerHttpAdapter.java
```diff
@@ -1,13 +1,24 @@
...
 import ratpack.zipkin.ServerResponse;
+import zipkin2.Endpoint;
 
 /**
  * This class is responsible for adapting Ratpack-specific request and responses
  * to something that brave.http.HttpServerParser can use to create the Span.
  */
 final class ServerHttpAdapter extends brave.http.HttpServerAdapter<ServerRequest, ServerResponse> {
+  @Override
+  public boolean parseClientAddress(final ServerRequest serverRequest, final Endpoint.Builder
```
codefromthecrypt: killme-nit: wonder if removing final qualifiers it will not wrap :P

src/test/java/brave/http/ITTracingFeature_Server.java
```diff
@@ -0,0 +1,40 @@
...
+
+import com.google.inject.Module;
+import com.google.inject.util.Modules;
+import ratpack.guice.Guice;
+import ratpack.test.embed.EmbeddedApp;
+import ratpack.zipkin.ServerTracingModule;
+
+import java.net.URI;
+
+public class ITTracingFeature_Server extends ITHttpServer {
```
codefromthecrypt, line 11: 👍

codefromthecrypt (approved):

src/main/java/ratpack/zipkin/internal/ServerHttpAdapter.java
hyleung: lol, I can just move the param to a new line and line them up...

PR #55, opened 2018-01-31
Uses latest brave to avoid an indirect dependency on io.zipkin.java

codefromthecrypt:
Technically, we still need a dep on io.zipkin.java:zipkin due to
overloading spanReporter. However, this change removes a hidden
dependency caused by Brave's overload on localEndpoint.

### DIFF
--- a/src/main/java/ratpack/zipkin/ServerTracingModule.java
+++ b/src/main/java/ratpack/zipkin/ServerTracingModule.java
@@ -23,7 +23,6 @@ import brave.http.HttpClientParser;
 import brave.http.HttpSampler;
 import brave.http.HttpServerParser;
 import brave.http.HttpTracing;
-import brave.internal.Platform;
 import brave.propagation.B3Propagation;
 import brave.propagation.Propagation;
 import brave.sampler.Sampler;
@@ -90,9 +89,8 @@ public class ServerTracingModule extends ConfigurableModule<ServerTracingModule.
     Tracing tracing = Tracing.newBuilder()
                              .sampler(config.sampler)
                              .currentTraceContext(new RatpackCurrentTraceContext())
-                             .localEndpoint(buildLocalEndpoint(config.serviceName, serverConfig.getPort(),
+                             .endpoint(buildEndpoint(config.serviceName, serverConfig.getPort(),
                                  serverConfig.getAddress()))
-                             .localServiceName(config.serviceName)
                              .spanReporter(config.spanReporter)
                              .propagationFactory(config.propagationFactory)
                              .build();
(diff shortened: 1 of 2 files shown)


src/main/java/ratpack/zipkin/ServerTracingModule.java
```diff
@@ -104,10 +102,11 @@ public HttpTracing getHttpTracing(final Config config, final ServerConfig server
                       .build();
   }
 
-  private static Endpoint buildLocalEndpoint(String serviceName, int port, @Nullable InetAddress configAddress) {
+  private static Endpoint buildEndpoint(String serviceName, int port, @Nullable InetAddress configAddress) {
     Endpoint.Builder builder = Endpoint.newBuilder();
     if (!builder.parseIp(configAddress)) {
-      builder = Platform.get().localEndpoint().toBuilder();
+      // TODO: shade brave.internal.Platform
```
hyleung, line 108: What does "shade" mean in this context? i.e. what needs to be done?

hyleung:
Looks like the server tracing integration tests are failing, will try too look into it this weekend!

codefromthecrypt:
sorry I mean repackage the type for example to ratpack.internal which makes
sure this doesn't end up breaking if brave changes internal types.

https://maven.apache.org/plugins/maven-shade-plugin/

hyleung:
Ok, finally got around to looking at this. Looks like https://github.com/openzipkin/brave/pull/580 added a couple of new test cases to `ITHttpServer` and `ITHttpClient` testing support for "extra" data propagation.

hyleung:
@adriancole so, I got it down to 2 failures on the extra propagation for the HTTP client (server stuff works fine). In a way it's not super surprising because the client tracing stuff is a bit kludgy because until recently there wasn't out-of-the-box support for client interceptors in Ratpack. I have some WIP that reworks the client tracing code using Ratpack's interceptor support (mostly works except for redirect handling).

Given that the currently client tracing code might (fingers crossed) be gone soon, I'm probably going to keep going on that and revisit this a little later. 

Oh, the branch that I've been working on for *this* stuff (updating to the latest Brave) is here: https://github.com/hyleung/ratpack-zipkin/tree/upgrade-brave, in case anybody is in dire need of the latest stuff and feels like taking it further.

codefromthecrypt:
Thanks for the heads up. I actually do want to collaborate. I have some
http.template work going on right now and I am testing with vert.x which
has a similar concern to ratpack (initial handler doesnt always know the
route yet)

Ps nasty bug in extrafieldpropagation is fixed in brave 4.15

codefromthecrypt:
FYI taking a look at that branch now

codefromthecrypt:
should be good now. waiting for 4.15.1 as I had a screwup in the client test fixture which broke assumptions about propagation handling here (causing one of the tests to fail)

codefromthecrypt:
obviated by https://github.com/hyleung/ratpack-zipkin/pull/57

PR #56, opened 2018-02-13
Fixes propagation handling

codefromthecrypt:
awaiting a release of brave 4.15.1 as it fixes a small glitch in the test fixture

### DIFF
--- a/src/main/java/ratpack/zipkin/internal/DefaultServerTracingHandler.java
+++ b/src/main/java/ratpack/zipkin/internal/DefaultServerTracingHandler.java
@@ -9,8 +9,6 @@ import brave.propagation.TraceContext;
 import javax.inject.Inject;
 
 import com.google.common.net.HostAndPort;
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
 import ratpack.handling.Context;
 import ratpack.handling.Handler;
 import ratpack.http.Headers;
@@ -20,7 +18,6 @@ import ratpack.http.Response;
 import ratpack.http.Status;
 import ratpack.path.PathBinding;
 import ratpack.server.PublicAddress;
-import ratpack.server.ServerConfig;
 import ratpack.zipkin.ServerRequest;
 import ratpack.zipkin.ServerResponse;
 import ratpack.zipkin.ServerTracingHandler;
@@ -35,7 +32,7 @@ public final class DefaultServerTracingHandler implements ServerTracingHandler {
   private final Tracing tracing;
   private final HttpServerHandler<ServerRequest, ServerResponse> handler;
   private final TraceContext.Extractor<ServerRequest> extractor;
-  private final Logger logger = LoggerFactory.getLogger(DefaultServerTracingHandler.class);
+
   @Inject
   public DefaultServerTracingHandler(final HttpTracing httpTracing) {
     this.tracing = httpTracing.tracing();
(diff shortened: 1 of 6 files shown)


codefromthecrypt:
ok bumped to released version and fingers crossed

src/main/java/ratpack/zipkin/internal/DefaultServerTracingHandler.java
```diff
@@ -48,15 +45,15 @@ public void handle(Context ctx) throws Exception {
...
-    final Tracer.SpanInScope scope = tracing.tracer().withSpanInScope(span);
     ctx.getResponse().beforeSend(response -> {
       ServerResponse serverResponse = new ServerResponseImpl(response, request, ctx.getPathBinding());
       handler.handleSend(serverResponse, null, span);
-      scope.close();
     });
-    ctx.next();
+    //place the Span in scope so that downstream code (e.g. Ratpack handlers
+    //further on in the chain) can see the Span.
+    try (Tracer.SpanInScope scope = tracing.tracer().withSpanInScope(span)) {
```
llinder, line 54: good catch!

llinder (approved):
Nice improvements!  Thanks a ton for the contributions.

src/main/java/ratpack/zipkin/internal/DefaultServerTracingHandler.java
codefromthecrypt: something yelled in the logs :) my goal with the base test class is to make it scream as much as possible, as frankly instrumentation is very hard job of remembering and eagle eyes otherwise! cc @jcchavezs

src/main/java/ratpack/zipkin/internal/ZipkinHttpClientImpl.java
```diff
@@ -43,13 +49,19 @@
...
+    private final ThreadLocalSpan span;
+    private final BiFunction<WrappedRequestSpec, TraceContext, Span> nextSpan;
     private final HttpClientHandler<WrappedRequestSpec, Integer> handler;
     private final TraceContext.Injector<MutableHeaders> injector;
 
     @Inject
     public ZipkinHttpClientImpl(final HttpClient delegate, final HttpTracing httpTracing) {
         this.delegate = delegate;
-        this.handler = HttpClientHandler.create(httpTracing, new HttpAdapter());
+        this.span = ThreadLocalSpan.create(httpTracing.tracing().tracer());
```
hyleung, line 61: On the usage of `ThreadLocalSpan`, looking at the Javadoc for that class, re. the callbacks:

```
Provided the library guarantees these run on the same thread, you can simply propagate the
 * result of {@link Tracer#withSpanInScope(Span)} from the starting callback to the closing one.
```
Don't know enough about how Ratpack does things to be able to verify that we have these guarantees - @llinder, do you know? I think in our case, the "onStart/onFinish" callbacks would be in the request method:

https://github.com/adriancole/ratpack-zipkin/blob/07f7e18f0f7d46dd6c83bcbca16bee74bc0694aa/src/main/java/ratpack/zipkin/internal/ZipkinHttpClientImpl.java#L98-L99

src/main/java/ratpack/zipkin/internal/DefaultServerTracingHandler.java
hyleung: I would never have spotted that - out of curiosity, what do the logs say in this case (and would it warrant a test failure)?

codefromthecrypt:
I *thought* I saw a message about a leak. I immediately looked for things
that could leak (span in scope is the only thing besides things that could
use it threadlocalspan).

It was a shotgun response

src/main/java/ratpack/zipkin/internal/ZipkinHttpClientImpl.java
llinder: This seems like it should work ok.  Since `ThreadLocalSpan` is deferring to the tracer and thus the configured `CurrentTraceContext`, it should still be on a Ratpack managed thread and thus the Registry values should be in a reliable state.

Then again there are a lot of "should"s in there :)  Without really digging into Ratpack code I can't say what will happen with 100% certainty though :)

The pattern of using a partial function with the ThreadLocalSpan tidies things up nicely though and is worth a try.

llinder (approved):

codefromthecrypt:
So the risk here is that there's a thread jump between the client
requestConfigurer and the wiretap listener. I'll poke around for a
different approach, and regardless add some code comments about the
impact.

codefromthecrypt:
last commit on https://github.com/hyleung/ratpack-zipkin/pull/57 fixes
the thread local approach. nice catch @hyleung

PR #57, opened 2018-02-15
Upgrades brave and adds current trace context support

codefromthecrypt:
This updates to the latest brave, and changes approach so that the current span can be read by user code (for example creating outgoing requests or using SpanCustomizer)

### DIFF
--- a/src/test/java/brave/http/ITZipkinHttpClientImpl.java
+++ b/src/test/java/brave/http/ITZipkinHttpClientImpl.java
@@ -15,7 +15,7 @@ import ratpack.util.Exceptions;
 import ratpack.zipkin.internal.RatpackCurrentTraceContext;
 import ratpack.zipkin.internal.ZipkinHttpClientImpl;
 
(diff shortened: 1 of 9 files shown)


src/test/java/brave/http/ITZipkinHttpClientImpl.java
```diff
@@ -15,7 +15,7 @@
 import ratpack.zipkin.internal.RatpackCurrentTraceContext;
 import ratpack.zipkin.internal.ZipkinHttpClientImpl;
 
-public class ITTracingFeature_Client extends ITHttpClient<HttpClient> {
+public class ITZipkinHttpClientImpl extends ITHttpAsyncClient<HttpClient> {
```
codefromthecrypt, line 18: @hyleung PS the base test classes here now have leak detection. So if you still have a span in scope when you go to close a span, a leak might have happened. This helped me notice we were never using the streaming variant in an integration test.

hyleung (approved):
Sweet! Thanks :)

PR #58, opened 2018-02-15
Updates to latest brave, supporting route-based span names

codefromthecrypt:
this tests the api design here https://github.com/openzipkin/brave/pull/602

### DIFF
--- a/README.md
+++ b/README.md
@@ -291,17 +291,17 @@ tracer.finishSpan();
 
 Currently nested local spans don't work in v1, but are supported in v2.
 
-#### Default KeyValueAnnotations
+#### Default Tags
 
 For server requests, we record the following annotations:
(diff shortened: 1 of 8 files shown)


hyleung:
I guess once this stuff is moved into the adapter, we can deprecate or remove kludgy stuff from `SpanNameProvider` 😁

https://github.com/hyleung/ratpack-zipkin/blob/master/src/main/java/ratpack/zipkin/SpanNameProvider.java#L11-L25

codefromthecrypt:
ok did the minimum here. We can decide whether or not to remove the span namer independently

codefromthecrypt:
going to bump for a patch release in a bit

codefromthecrypt:
ok ready to go. tested with the example app

screen shot 2018-02-23 at 10 25 36 pm

hyleung (approved):
Thanks!

PR #61, opened 2018-05-22
Removes old dependencies via Brave 5

codefromthecrypt:
Brave v5 is the same as v4 except no old deps. Notably, there is no
current constant file, so that's why literals are used in this change.

See https://github.com/openzipkin/brave/issues/699

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -35,15 +35,12 @@ repositories {
 }
 
 dependencies {
-    implementation "io.zipkin.brave:brave-bom:4.18.2"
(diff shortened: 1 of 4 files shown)


hyleung (approved):
Lemme know if you think releasing a new version is necessary. I'm just starting to get caught up (up to my eyeballs in GDPR for the past 5 months).

codefromthecrypt:
forgot we didn't merge this.. bumped to latest version of brave 5 and will merge on re-green

PR #62, opened 2018-05-26
Upgrade to Ratpack v1.5.4

hyleung:
Originally I started doing this to resume work on the Http client, but realized that the changes in Ratpack that I want to use haven't actually been released yet. In any case, figured it wouldn't hurt to bump the version of Ratpack to the current release version anyway.


src/main/java/ratpack/zipkin/internal/ZipkinHttpClientImpl.java
```diff
@@ -251,6 +269,11 @@ public RequestSpec maxContentLength(int numBytes) {
             return this;
         }
 
+        @Override
+        public RequestSpec responseMaxChunkSize(final int numBytes) {
+            return null;
```
codefromthecrypt: oops this :P

hyleung: 🤦‍♂️

PR #67, opened 2018-12-10
WIP - Upgrades Ratpack 1.6 and Brave 5.5.1

llinder:
- replaces client instrumentation wrapper with new interceptor api.
- updated spock tests for ratpack 1.6


pom.xml
```diff
@@ -41,8 +41,8 @@
     <maven-install-plugin.version>2.5.2</maven-install-plugin.version>
     <maven-deploy-plugin.version>2.8.2</maven-deploy-plugin.version>
 
-    <braveVersion>5.1.5</braveVersion>
-    <ratpackVersion>1.4.6</ratpackVersion>
+    <braveVersion>5.5.1</braveVersion>
+    <ratpackVersion>1.6.0-rc-2</ratpackVersion>
```
llinder, line 45: 1.6 hasn't been released yet so we should wait for that.

src/main/java/ratpack/zipkin/ServerTracingModule.java
```diff
@@ -55,11 +55,28 @@ protected void configure() {
...
+
+    Provider<HttpClient> httpClientProvider = () ->
+            // HttpClient.of throws exceptions but Providers can not, hence the unchecked wrapper.
+            Exceptions.uncheck(() -> HttpClient.of(spec -> {
+              spec.requestIntercept(clientTracingInterceptorProvider.get()::request);
+              spec.responseIntercept(clientTracingInterceptorProvider.get()::response);
+            }));
 
-    bind(ZipkinHttpClientImpl.class);
+    bind(HttpClient.class)
```
llinder, line 70: This changes behavior in that all bindings on HttpClient will now use one that is configured for tracing.  This may be a confusing change if folks are use to the required `@Zipkin` annotation.  In practice however, I have found it more likely that folks forget the `@Zipkin` annotation and wonder why there is no client tracing.

src/main/java/ratpack/zipkin/internal/DefaultClientTracingInterceptor.java
```diff
@@ -0,0 +1,59 @@
...
+
+import javax.inject.Inject;
+
+public class DefaultClientTracingInterceptor implements ClientTracingInterceptor {
+
+  private static final TypeToken<Span> SpanToken = new TypeToken<Span>() {};
+
+  private final HttpClientHandler<RequestSpec, Integer> handler;
+  private final TraceContext.Injector<MutableHeaders> injector;
+  private final Execution execution;
```
llinder, line 24: According to feedback I received from the Ratpack Slack channel, we should be able to rely on the Ratpack Execution to maintain trace context state for us.  While the tests do pass with this configuration, we should still try it on a real application just to be sure.

src/test/java/brave/http/ITZipkinHttpClient.java
```diff
@@ -0,0 +1,93 @@
...
+  }
+
+  @Override protected void post(HttpClient client, String pathIncludingQuery, String body) throws Exception {
+    harness.yield(e ->
+      client.post(URI.create(url(pathIncludingQuery)), (request ->
+        request.body(b -> b.text(body))
+      ))
+    ).getValue();
+    // small delay to ensure the response interceptor is invoked before this method exits.
+    Thread.sleep(1);
```
llinder, line 82: I'm not sure why but the response interceptor ran after the blocking get call.  While not a thrilling solution, adding a sleep here fixed the issue.  Since the tests are written in a blocking fashion, and Ratpack code is written in a non blocking fashion, I suspect that actual production code will work as expected.  Again we should investigate testing this more... maybe adding some Ratpack specific tests is warranted here before merging.

src/main/java/ratpack/zipkin/ServerTracingModule.java
beckje01: I think this is inline with how tracing should work.

beckje01:
Given the major changes I think it would make sense to rev the major version on this

